Issue #8, opened 2018-05-22
Add df() method to Bars


umitanuki:
As a python algo writer, I want a df() method that returns pandas.DataFrame from Bars entity.  It should not be hard since pandas are already part of the dependency

PR #11, opened 2018-05-27
Official support of `df` property and CI support

umitanuki:


umitanuki:
#6 and #8 are addressed

PR #14, opened 2018-06-12
Fixing readme typo

raovinnakota:
In the installation instructions, the api is referred to as alpaca-trade-api-python, but when using `pip search`, the only package with that name found was alpaca-trade-api


umitanuki:
Please rebase to the latest master

PR #22, opened 2018-07-19
staging support

smartchris84:


alpaca_trade_api/stream2.py
```diff
@@ -89,7 +90,8 @@ def run(self, initial_channels=[]):
         '''Run forever and block until exception is rasised.
         initial_channels is the channels to start with.
         '''
-        loop = asyncio.get_event_loop()
+        loop = asyncio.new_event_loop()
```
umitanuki: Hmm I don't think this works in the normal case aka main thread.

umitanuki:
Thanks for the proposal.  I've incorporated the change in #23

Issue #34, opened 2018-10-24
Add missing parameters


umitanuki:
http://docs.alpaca.markets/web-api/orders/#get-a-list-of-orders

Add parameters that are now documented:

- limit
- after
- until
- direction

PR #35, opened 2018-10-27
fixes: #34 list_orders takes params

h55nick:
allow `list_orders` to take a `params` dict that will be passed through to `.get` as query parameters. 

Note: No validation of valid params here but should be done at the API level

### DIFF
--- a/alpaca_trade_api/rest.py
+++ b/alpaca_trade_api/rest.py
@@ -130,11 +130,22 @@ class REST(object):
         resp = self.get('/account')
         return Account(resp)
 
-    def list_orders(self, params=dict(), status=None):
+    def list_orders(self, status=None, limit=None, after=None, until=None,
+                    direction=None, params=None):
         '''
         Get a list of orders
         https://docs.alpaca.markets/web-api/orders/#get-a-list-of-orders
         '''
+        if params is None:
+            params = dict()
+        if limit is not None:
+            params['limit'] = limit
+        if after is not None:
+            params['after'] = after
+        if until is not None:
+            params['until'] = until
+        if direction is not None:
+            params['direction'] = direction
         if status is not None:
             params['status'] = status
         resp = self.get('/orders', params)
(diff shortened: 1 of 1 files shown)


h55nick:
also considering changing:
```
        params['status'] = params['status'] or status or None #might not be very pythonic
        resp = self.get('/orders', params)
        return [Order(o) for o in resp]
```
to simplify

umitanuki:
Thanks for the PR. I can see bundling parameters into `param` may be clean, but for now I would rather keep them separate for reasons: 1) to be consistent with other methods, and 2) documentation/explanation purpose (you can do rest.list_orders? in ipython to see parameter names quickly)

alpaca_trade_api/rest.py
```diff
@@ -130,9 +130,11 @@ def get_account(self):
         resp = self.get('/account')
         return Account(resp)
 
-    def list_orders(self, status=None):
-        '''Get a list of orders'''
-        params = dict()
+    def list_orders(self, params=dict(), status=None):
```
umitanuki: FYI, `dict()` is initialized and reused this way, which means the second invocation will have the same dict instance with keys filled as in the first call.
https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

h55nick: fascinating, certainly explains it well and curious why I was never aware! Can change once we decide on the direction.

h55nick:
@umitanuki understand the desire to be clean of course. If you have an example of passing through parameters I'd be happy to make the update. One of my goals here would be not to require consistent updates to this repo for each api update so instead of having it strictly pass-through vs containing logic. This is why I would like to actively avoid adding parameters as 'status' is used now (kept status so we didn't break backward compatibility). 

Many thanks for your comments! Let me know what you are thinking.

h55nick:
@umitanuki I've updated the PR according to your comments but leaving the option to pass in parameters like done previously. Let me know what you think :-)

umitanuki:
Looks good to me. Thanks for the work!

PR #38, opened 2018-11-21
denied redirects on requests

ttt733:


alpaca_trade_api/rest.py
```diff
@@ -69,6 +69,7 @@ def _request(self, method, path, data=None, prefix='/v1'):
         }
         opts = {
             'headers': headers,
+            'allow_redirects': False
```
umitanuki: Can you add the reason in a comment?

umitanuki:
Looks good

PR #39, opened 2018-11-27
Add accessor for new bars data endpoint

ttt733:


alpaca_trade_api/entity.py
```diff
@@ -23,6 +23,9 @@ def __getattr__(self, key):
                      key.endswith('_time')) and
                     ISO8601YMD.match(val)):
                 return pd.Timestamp(val)
+            elif (isinstance(val, int) and
+                    key is 't'):
```
umitanuki: `is` isn't very appropriate here?

alpaca_trade_api/entity.py
```diff
@@ -23,6 +23,9 @@ def __getattr__(self, key):
...
+                return pd.Timestamp(val)
```
umitanuki: That's not quite right. "t" is in seconds

PR #41, opened 2018-12-05
Converts bytes to str before calling json.loads()

qyuga:
It fixes the following error while calling json.loads() in some Python versions including 3.5.6:
TypeError: the JSON object must be str, not 'bytes'


ttt733:
Pull request #42 addresses this along with another instance of the issue.

umitanuki:
@qyuga thanks for the fix!

PR #42, opened 2018-12-05
Fixed JSON str vs bytes error:

Devination:
`TypeError: the JSON object must be str, not 'bytes'`


umitanuki:
@Devination Thanks for the fix!

PR #45, opened 2018-12-13
Fix dataframe conversion when Bars data is empty

ttt733:


alpaca_trade_api/entity.py
```diff
@@ -80,9 +80,14 @@ def df(self):
             }
             df.columns = [alias[c] for c in df.columns]
             df.set_index('time', inplace=True)
-            df.index = pd.to_datetime(
-                df.index * 1e9, utc=True,
-            ).tz_convert(NY)
+            if not df.empty:
```
umitanuki: What fails if it's empty?

ttt733: tz.convert

PR #46, opened 2018-12-18
Added POLYGON_API_KEY_ID as optional distinct credential

bdowling:
I thought it might be useful to have the ability to independently set the polygon API key_id.  In the current environment, this can be useful if a funded account holder wants to use the paper trading API with polygon (paper keys do not currently work with polygon NATs).

Also updated the readme to provide a little more clarity on passing this and the other keys to the API.

Note: this PR also adds a bit of checking for the websocket API credential check, raising an error if it is not successful.  This does not address the polygon auth errors.


umitanuki:
Paper-only account does not work with polygon, but paper account for the user who has live account should work with polygon.

umitanuki:
This change would break existing uses, since most of the users have no problem with Alpaca key id for polygon.

bdowling:
@umitanuki I tried to make it as backwards compatible as possible, the POLYGON key ID is optional, it will default to using the provided APAC_API_KEY_ID, providing preference to the kwarg, environment, then default.

PR #47, opened 2018-12-18
Added A.* polygon stream as Agg entity with added mappings

bdowling:
Adds support for the Polygon A.* Aggregation and mappings for some of the additional fields seen in that stream.


umitanuki:
Thanks for the patch!

PR #60, opened 2019-03-15
Add option to cancel an order by client order id

### LINKED ISSUE #59: Request ability to cancel order by client order id
If we are able to tag our orders with a Client Order ID, it seems reasonable for a call like `api.cancel_order_by_client_order_id('my_custom_client_order_id')` to exist. I ask because it seems rather confusing to provide developers with a nice way of tagging orders internally, then forcing them to query for the proper id with a call to `get_order_by_client_order_id` if they chose to use that feature.

ztaylor54:
Resolves #59 with a simple def wrapping a call to `get_order_by_client_order_id`.

### DIFF
--- a/alpaca_trade_api/rest.py
+++ b/alpaca_trade_api/rest.py
@@ -207,6 +207,11 @@ class REST(object):
         '''Cancel an order'''
         self.delete('/orders/{}'.format(order_id))
 
+    def cancel_order_by_client_order_id(self, client_order_id):
+        '''Cancel an order by client order id'''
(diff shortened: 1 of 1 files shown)


umitanuki:
I understand this is a workaround on the client side, but does it make sense to support it at API layer?

ttt733:
I agree with Hitoshi here - this seems like something that should be handled by the API rather than the client. I'll go ahead and file an issue on our alpacahq/Alpaca-API repository and see if we can't get this implemented on the backend.

PR #73, opened 2019-05-31
Detect and mitigate websockets disconnect.

aprsa:
A new task, test_connection(), added to the main event loop that catches the disconnect state of websockets by flushing the logger output to a string buffer. This allows for automatic reconnects in the case of ws failure.

### DIFF
--- a/alpaca_trade_api/stream2.py
+++ b/alpaca_trade_api/stream2.py
@@ -2,6 +2,8 @@ import asyncio
 import json
 import re
 import websockets
+import logging
+import io
 from .common import get_base_url, get_credentials
 from .entity import Account, Entity
 from . import polygon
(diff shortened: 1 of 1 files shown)


alpaca_trade_api/stream2.py
```diff
@@ -98,12 +100,44 @@ def __init__(self, key_id=None, secret_key=None, base_url=None):
...
+
+        while True:
+            await asyncio.sleep(poll_delay)
+            buffer_text = buffer.getvalue()
+            if 'CLOSED' in buffer_text:
+                print('stream closed')
+                while True:
+                    try:
+                        print('attempting to reconnect...', end='\r')
+                        # self._ws = await self._connect()
```
ttt733: This line can be deleted, right?

alpaca_trade_api/stream2.py
```diff
@@ -98,12 +100,44 @@ def __init__(self, key_id=None, secret_key=None, base_url=None):
...
+        logger.addHandler(logging.StreamHandler(stream=buffer))
+
+        while True:
+            await asyncio.sleep(poll_delay)
+            buffer_text = buffer.getvalue()
+            if 'CLOSED' in buffer_text:
+                print('stream closed')
+                while True:
+                    try:
+                        print('attempting to reconnect...', end='\r')
```
ttt733: Is it possible to use the logger rather than prints for this kind of information? I would rather not print anything using print() - if we can't push it to the logger, I'd rather just not print anything and let exceptions bubble up when reconnection fails.

alpaca_trade_api/stream2.py
```diff
@@ -98,12 +100,44 @@ def __init__(self, key_id=None, secret_key=None, base_url=None):
...
+                        # self._ws = await self._connect()
+                        self._ws = None
+                        await self.subscribe(initial_channels)
+                        break
+                    except:
+                        await asyncio.sleep(reconnect_delay)
+                        pass
```
ttt733: I think pass can be deleted here.

alpaca_trade_api/stream2.py
aprsa: Yep, that was my original attempt that didn't pan out. It's not enough to reconnect, we also need to re-subscribe.

alpaca_trade_api/stream2.py
aprsa: I would drop the print statements altogether in production code; they are there only to demonstrate what is going on at this PR stage.

alpaca_trade_api/stream2.py
aprsa: Yep, it was just a placeholder to figure out the reconnect_delay part. Sorry about that.

ttt733:
We've incorporated these sorts of changes into a recent release via a separate PR. Thank you for the idea!

Issue #64, opened 2019-04-08
Websocket exception


carles-estevadeordal:
The websocket sometimes gives an exception but it is not clear if it reconnects...

Task exception was never retrieved
future: <Task finished coro=<StreamConn._consume_msg() done, defined at /opt/conda/lib/python3.6/site-packages/alpaca_trade_api/stream2.py:46> exception=ConnectionClosed('WebSocket connection is closed: code = 1006 (connection closed abnormally [internal]), no reason',)>
Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/site-packages/websockets/protocol.py", line 674, in transfer_data
    message = yield from self.read_message()
  File "/opt/conda/lib/python3.6/site-packages/websockets/protocol.py", line 742, in read_message
    frame = yield from self.read_data_frame(max_size=self.max_size)
  File "/opt/conda/lib/python3.6/site-packages/websockets/protocol.py", line 815, in read_data_frame
    frame = yield from self.read_frame(max_size)
  File "/opt/conda/lib/python3.6/site-packages/websockets/protocol.py", line 884, in read_frame
    extensions=self.extensions,
  File "/opt/conda/lib/python3.6/site-packages/websockets/framing.py", line 99, in read
    data = yield from reader(2)
  File "/opt/conda/lib/python3.6/asyncio/streams.py", line 674, in readexactly
    yield from self._wait_for_data('readexactly')
  File "/opt/conda/lib/python3.6/asyncio/streams.py", line 464, in _wait_for_data
    yield from self._waiter
concurrent.futures._base.CancelledError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/site-packages/alpaca_trade_api/stream2.py", line 50, in _consume_msg
    r = await ws.recv()
  File "/opt/conda/lib/python3.6/site-packages/websockets/protocol.py", line 434, in recv
    yield from self.ensure_open()
  File "/opt/conda/lib/python3.6/site-packages/websockets/protocol.py", line 658, in ensure_open
    ) from self.transfer_data_exc
websockets.exceptions.ConnectionClosed: WebSocket connection is closed: code = 1006 (connection closed abnormally [internal]), no reason

aprsa:
Same issue here; it does not reconnect as no new messages come through the stream. Alpaca never passes the exception. Could it be that the websocket doesn't raise an exception so that alpaca api cannot catch it?

[...]

PR #88, opened 2019-07-10
Add reconnection handling into Polygon WS

### LINKED ISSUE #61: Stream auto-reconnect or notify on close
Is there a way to have streams auto-reconnect or register for a notification when the connection is closed?

bdowling:
This PR is an attempt to add some re-connection logic into the polygon websockets interface.  Currently it will use the same ENV variables as the REST api for retry count and delay if connections keep failing.

This PR also introduces two synthetic events that can be seen by the caller in the status event handler, an example re-connection trace looks like:
```
Entity({'ev': 'status', 'message': 'subscribed to: A.*', 'status': 'success'})
Entity({   'ev': 'status',
    'message': 'Polygon Disconnected Unexpectedly',
    'status': 'disconnected'})
Entity({'ev': 'status', 'message': 'Connecting to Polygon', 'status': 'connecting'})
Entity({'ev': 'status', 'message': 'Connecting to Polygon', 'status': 'connecting'})
Entity({'ev': 'status', 'message': 'Connected Successfully', 'status': 'connected'})
Entity({'ev': 'status', 'message': 'authenticated', 'status': 'success'})
```

Bumps websocket dependency to 8.0 which was just released that includes improvements in the exceptions it generates.

Replaces #73, Closes #61, #64

### DIFF
--- a/alpaca_trade_api/polygon/stream2.py
+++ b/alpaca_trade_api/polygon/stream2.py
@@ -1,5 +1,6 @@
 import asyncio
 import json
+import time
 import re
 import os
 import websockets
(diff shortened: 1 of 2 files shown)


aprsa:
Note that this does /not/ replace #73 -- #73 addressed trade update disconnects.

PR #99, opened 2019-07-26
Polygon streaming _handlers reference fix

eliNorth:
Fixed both StreamConn._handlers and StreamConn.polygon._handlers referencing the same object and thus causing KeyErrors on StreamConn.deregister() by shallow cloning the _handlers dictionary in StreamConn._ensure_polygon().


ttt733:
Looks good, thanks!

PR #103, opened 2019-08-06
Add example MartingaleTrader script

ttt733:


examples/martingale.py
```diff
@@ -0,0 +1,183 @@
...
+        # Figure out how much money we have to work with, accounting for margin
+        account_info = self.api.get_account()
+        self.equity = float(account_info.equity)
+        self.margin_multiplier = float(account_info.multiplier)
+        total_buying_power = self.margin_multiplier * self.equity
+        print(f'Initial total buying power = {total_buying_power}')
+
+
+    def start_trading(self):
+        conn = tradeapi.StreamConn(
```
andrewkim316: Isnt trade_updates unauthorized with paper trading api keys?

ttt733: No - you might be getting it confused with the issue where paper keys are incorrectly accepted by the live API.

examples/martingale.py
```diff
@@ -0,0 +1,183 @@
...
+                        self.streak_increasing = increasing
+                        self.streak_start = tick_open
+                        self.streak_count = 0
+                    else:
+                        # Calculate the number of shares we want to be holding
+                        total_buying_power = self.equity * \
+                            self.margin_multiplier
+                        target_value = (2**self.streak_count) * \
+                            (self.base_bet / 100) * total_buying_power
+                        if target_value > total_buying_power:
```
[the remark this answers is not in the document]
ttt733: Without removing one share from the final quantity, sometimes it was trying to order one more share than it could. I think it might have had to do with a discrepancy between this script's view of the price and Alpaca's view. Either way, I'd rather just set the max quantity it can order as one less than what it thinks is possible than dig too much into it.

examples/martingale.py
andrewkim316: Ah that might be it

examples/martingale.py
```diff
@@ -0,0 +1,185 @@
...
+                    self.current_order = None
+            elif event_type != 'new':
+                print(f'Unexpected order event type {event_type} received')
+
+        conn.run([f'A.{self.symbol}', 'trade_updates'])
+
+    def send_order(self, target_qty):
+        # We don't want to have two orders open at once
+        if self.current_order is not None:
+            self.api.cancel_order(self.current_order.id)
```
umitanuki: `self.current_order = None`

PR #108, opened 2019-08-17
Reconnect for websockets disconnect for non-polygon streams

aprsa:
This PR uses the new websockets 8.0.2 exceptions to catch a disconnect of the trade_updates stream and attempt to reconnect to the stream. It tries to reconnect every _reconnect_interval seconds, by default 5s.


ttt733:
I went ahead and implemented something that accomplishes this and fixes a few other issues in #111.

PR #110, opened 2019-08-19
Add OAuth authentication method

ttt733:
Alpaca will be providing limited support for OAuth authentication in the future, so this PR adds it as an alternative method of authentication.

### DIFF
--- a/alpaca_trade_api/common.py
+++ b/alpaca_trade_api/common.py
@@ -12,7 +12,7 @@ def get_data_url():
 
 
 def get_credentials(key_id=None, secret_key=None, oauth=None):
-    oauth = oauth or os.environ.get('APCA_API_OAUTH')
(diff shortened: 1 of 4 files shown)


alpaca_trade_api/common.py
```diff
@@ -11,23 +11,25 @@ def get_data_url():
         'APCA_API_DATA_URL', 'https://data.alpaca.markets').rstrip('/')
 
 
-def get_credentials(key_id=None, secret_key=None):
+def get_credentials(key_id=None, secret_key=None, oauth=None):
+    oauth = oauth or os.environ.get('APCA_API_OAUTH')
```
umitanuki: OAUTH_TOKEN maybe? @Neal opinion?

Neal: I agree, OAUTH_TOKEN sounds better. Maybe also prefix it with APCA?

PR #111, opened 2019-08-20
Add automatic reconnection for Alpaca stream

ttt733:
Though the Alpaca API itself rarely closes websocket connections, local network conditions have caused some users to report abnormal connection closures. This should address some of the issues people have been seeing as a result.

### DIFF
--- a/alpaca_trade_api/polygon/streamconn.py
+++ b/alpaca_trade_api/polygon/streamconn.py
@@ -121,13 +121,12 @@ class StreamConn(object):
                 await self.connect()
                 if self._streams:
                     await self.subscribe(self._streams)
-
                 break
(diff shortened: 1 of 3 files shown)


alpaca_trade_api/polygon/streamconn.py
```diff
@@ -121,13 +121,12 @@ def __init__(self, key_id=None):
                 await self.connect()
                 if self._streams:
                     await self.subscribe(self._streams)
-
                 break
-            except (ConnectionRefusedError, ConnectionError) as e:
+            except Exception as e:
```
ttt733: I had to make this less specific because in some cases, such as wifi being disabled on the machine running a script, unusual errors are the ones which reach this code.

PR #118, opened 2019-10-13
Add previous_day_bar

Devination:


ttt733:
Hi @Devination , thanks for the PR! Could you please change this so that it creates it as an Aggv2 rather than a Bar object? Bars are intended for results from the Alpaca API, while Aggs represent those from the Polygon API.

Devination:
Switched it to Aggsv2, good call Trevor!

PR #120, opened 2019-10-28
question regarding threading module usage

popeliao:


ttt733:
I believe the threading module is just what handles joining the threads.

PR #121, opened 2019-10-28
Incorrect bar timeframe

popeliao:
Comment says 10 days, but it is actually getting 10 minutes

Also the bar endpoint documentation is very confusing.
https://docs.alpaca.markets/api-documentation/api-v2/market-data/bars/
It doesn't specify what is the behavior w/o start, end, after, until.

I also try with a query with start=2019-10-25T09:30:00 end=2019-10-25T16:00:00, but the first bar is for October 25, 2019 1:34:00 PM
The last is bar is starting from October 25, 2019 3:59:00 PM
There are 100 bars in total. However I am not sure how these 100 minute bar is distributed around an about 2 and a half hour time span.

### DIFF
--- a/examples/long-short.py
+++ b/examples/long-short.py
@@ -309,7 +309,7 @@ class LongShort:
   def getPercentChanges(self):
     length = 10
     for i, stock in enumerate(self.allStocks):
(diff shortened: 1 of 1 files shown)


ttt733:
The code is correct - the comment is what needs to be updated.

PR #128, opened 2019-12-12
First Impression Solved : Add Base Url URL at REST Module

loaiabdalslam:
JupyterLab (2)

my first impression about package it was : why that error at just start after that i checked Issue's, i fount that i must add `https://paper-api.alpaca.markets` as Third Parameter Like Next Code 
While in The Docs you didn't Explain this point at that quick start

That Code Case Error :-1: 
```
import alpaca_trade_api as tradeapi

api = tradeapi.REST('<key_id>', '<secret_key>', api_version='v2') # or use ENV Vars shown below
account = api.get_account()
api.list_positions()
```

That Code Works Fine :+1: 
```
import alpaca_trade_api as tradeapi

api = tradeapi.REST('<key_id>', '<secret_key>','https://paper-api.alpaca.markets', api_version='v2') # or use ENV Vars shown below
account = api.get_account()
api.list_positions()
```

So i Added `'https://paper-api.alpaca.markets'` as Default Parameter if we Pass None .

### DIFF
--- a/alpaca_trade_api/rest.py
+++ b/alpaca_trade_api/rest.py
@@ -58,7 +58,7 @@ class REST(object):
         self,
         key_id=None,
         secret_key=None,
-        base_url=None,
+        base_url='https://paper-api.alpaca.markets',
         api_version=None,
         oauth=None
     ):


ttt733:
The live API being the default when none is specified is intentional. Unfortunately, this would be a breaking change for users using the SDK for live trading who expect to not have to specify the URL. The readmes and docs show how to use environment variables in such a way that you don't need to specify the paper endpoint in the code, as well.

PR #133, opened 2019-12-31
Update all examples to have api key filler

martell:
This brings these 2 examples inline with the others


umitanuki:
In a way, not having placeholder variables is also a valid and I actually prefer, since it is runnable without code modification, if you plug in environment variables. These placeholders force to be updated first to run the code.

PR #138, opened 2020-01-14
Added Alpha Vantage Integration

PatrickAlphaVantage:


PatrickAlphaVantage:
Hey guys, speaking with Hitoshi he told me that some of the Alpaca users would love some type of integration with Alpha Vantage, and the best way to go would probably be an open source combination here. I spent a couple days blasting this out, my skills in your other SDKs are a little weak at the moment, but if there is a lot of interest I can look into making those too. Let me know what you think!

ttt733:
Hey, thanks for the PR! I'm going to merge this as-is, but I'm going to remove the forex and crypto functionality in a separate commit before releasing it. We do the same for our Polygon integration, since users can't trade forex or crypto through Alpaca. If we start offering that service at some point, we'll be able to add it back in.

PatrickAlphaVantage:
Nice! Thanks! If there is sufficient response to this, let me know and I can add some integrations to the other SDKs.

PR #150, opened 2020-02-18
Add Polygon end-pt: Daily Open/Close

t-shah:
This PR adds support for the following Polygon endpoint:
[https://polygon.io/docs/#!/Stocks--Equities/get_v1_open_close_symbol_date](https://polygon.io/docs/#!/Stocks--Equities/get_v1_open_close_symbol_date)

Current (Polygon) methods supported in the API return daily close data from 1 day prior, but not from the same day as requested.


ttt733:
I think if we're going to release an endpoint for this we'll want to also add an entity object to represent the response. If you want to add that to this PR, you can, if not I'll merge this and add it in a separate PR before the next release.

PR #152, opened 2020-02-21
FIX: 'from' keyword argument in polygon api docs

abpwrs:


ttt733:
The method uses `_from`. `_from` is used rather than `from` because `from` is a reserved keyword (used for imports.)

ttt733:
Misread this at first. Thanks!

PR #164, opened 2020-03-08
bugfix specify seconds as timestamp unit

stin7:
API returns timestamp in seconds, pandas to_datetime() has a default of ns https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_datetime.html


umitanuki:
Thanks for the quick find!

ttt733:
Sorry, just checked this out. Handling of the timestamp is already managed when we multiply the index by 1e9 - that effectively converts to nanoseconds already. If you're seeing a specific issue, open a GitHub issue and we can check it out - but in my testing, the timestamp column of the dataframe is accurate.

PR #168, opened 2020-03-22
Added StockTwits integration

bbeale:
Added code (and unit tests) to get data from StockTwits about given ticker symbols. I only implemented a few query endpoints that are relevant to stocks. Planning to eventually implement more endpoints, especially if there is interest. None of these endpoints require authentication, but users can get a StockTwits API key for free.

Endpoints included:
[streams/symbol](https://api.stocktwits.com/developers/docs/api#streams-symbol-docs)
[streams/charts](https://api.stocktwits.com/developers/docs/api#streams-charts-docs)
[streams/trending](https://api.stocktwits.com/developers/docs/api#streams-trending-docs)
[search/symbols](https://api.stocktwits.com/developers/docs/api#search-symbols-docs)
[trending/symbols/equities](https://api.stocktwits.com/developers/docs/api#trending-symbols-equities-docs)


umitanuki:
That's a very interesting integration. But I do have a question; why you want it in this SDK?

umitanuki:
Including other packages under this, I am also thinking about more pluggable approach. I'd welcome any idea.

bbeale:
I have been using it alongside of the market data endpoints and thought others might find it useful. I was intrigued back when I saw the Alpha Vantage integration get merged in, and that gave me a similar idea for StockTwits.

umitanuki:
I see where you are coming from. Having more plugin-like approach would help others get something useful easily but I want to make sure those are visible in the ecosystem.  Backtrader for example has this style and Alpaca has external repository (https://github.com/alpacahq/alpaca-backtrader-api). Python makes it easy to dynamically load package as well. Then maybe we can put it in the documentation. We put alpha vantage but it's not well documented either at the moment and I've been wondering what would be the best way for that too.

bbeale:
So I took a look at the Backtrader code the other day to see how StockTwits and AlphaVantage might fit in there. I agree that external repositories make sense, and we could probably just move those whole directories as they are with minimal trouble. 

Do you think it makes more sense for each integration to have it's own repo, i.e. `alpacahq/stocktwits`, `alpacahq/alpha-vantage`, or to have one repo -- `alpacahq/extensions` or whatever it would be called, with subdirectories `alpacahq/extensions/stocktwits`, `alpacahq/extensions/alpha-vantage`?

PR #170, opened 2020-03-25
Use v2 by default

shlomikushchi:
* use the V2 url path from https://polygon.io/docs
* use dateutil to format the dates properly
* documented historic_agg_v2()


umitanuki:
While this is desiable, as the change is breaking, we will do it but not in hurry. The code itself looks good to me.

PR #173, opened 2020-03-27
Use v2 by default for historic_agg_v2

shlomikushchi:
* use the V2 url path from https://polygon.io/docs
* use dateutil to format the dates properly
* documented historic_agg_v2()
* removed historic_agg() (v1) to avoid confusion.
* removed tests for historic_agg()


umitanuki:
That would break other methods. Can you rather remove v1 historic agg to avoid confusion?

shlomikushchi:
* also renoved historic_agg() (v1) to avoid confusion.
* removed tests for historic_agg()
* changed default to v1 in the rest.get()  method
* updated version in init file (0.47)

alpaca_trade_api/polygon/rest.py
```diff
@@ -109,27 +110,28 @@ def historic_quotes_v2(
...
+               e.g if 1 we get for daily 2015-01-05, 2015-01-06, 2015-01-07, 2015-01-08
+                   if 3 we get           2015-01-01, 2015-01-04, 2015-01-07, 2015-01-10)
+        :param timespan: Size of the time window: minute, hour, day, week, month, quarter, year
+        :param _from:
+        :param to:
+        :param unadjusted:
+        :param limit: max samples to retrieve (seems like we get "limit - 1" )
+        :return:
+        """
+        path = '/aggs/ticker/{symbol}/range/{multiplier}/{timespan}/{_from}/{to}'.format(symbol=symbol,
```
shlomikushchi, line 128: please note that some of our code that calls this method (e.g pylivetrader)
is calling it like this:
```py
self._api.polygon.historic_agg_v2(
                symbol, 1, size,
                int(_from.timestamp()) * 1000,       <====================
                int(to.timestamp()) * 1000      <====================
            )
```
and it doesn't work because this code expects a string.

so we should either fix all calling code to use the same format, or make this code adjustable to this scenario.

umitanuki:
@shlomikushchi so this is backward compatible?

PR #174, opened 2020-03-27
Entity float attribute should be float

shlomikushchi:
rest api returns data as string.
 we already transform datetime objects to Timestamp. 
let's also do that for floats


umitanuki:
No, it's by design. The type is actually decimal, not float, and it's up to the caller if wants to cast it to float.

PR #177, opened 2020-04-09
Fix Runtine error when trying to get event loop

shlomikushchi:
asyncio.get_event_loop() throws Runtime error because event loop doesn't exist in thread
we need to add that to the available exceptions that we create a new event loop when we get them

### DIFF
--- a/alpaca_trade_api/stream2.py
+++ b/alpaca_trade_api/stream2.py
@@ -25,7 +25,7 @@ class StreamConn(object):
         self.polygon = None
         try:
             self.loop = asyncio.get_event_loop()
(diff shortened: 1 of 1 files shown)


umitanuki:
Is this an exception that needs to be handled by SDK? If this is more programming error by the caller, should it be handled elsewhere?

shlomikushchi:
> Is this an exception that needs to be handled by SDK? If this is more programming error by the caller, should it be handled elsewhere?

I resolved it in the calling pakcage(backtrader) in this PR: https://github.com/alpacahq/alpaca-backtrader-api/pull/36 

closing this PR

PR #178, opened 2020-04-10
Support both the data and the trading stream simultaneously

gnvk:


alpaca_trade_api/stream2.py
```diff
@@ -103,51 +86,113 @@ def __init__(self, key_id=None, secret_key=None, base_url=None):
...
             else:
-                ws_channels.append(c)
+                data_channels.append(c)
 
-        if len(ws_channels) > 0:
-            # Currently our streams don't support unsubscribe
-            # not as useful with our feeds
-            pass
+        if trading_channels:
+            await self._ensure_ws(self.trading_ws)
```
umitanuki, line 180: I would prefer to make 3 channels async to each other. As it stands, this is sequential.

umitanuki: Hmm but it is what it was, so maybe it's ok.

alpaca_trade_api/stream2.py
```diff
@@ -103,51 +86,113 @@ async def _ensure_ws(self):
...
+
+    def __init__(self, key_id=None, secret_key=None, base_url=None, data_url=None):
+        _key_id, _secret_key, _ = get_credentials(key_id, secret_key)
+        _base_url = base_url or get_base_url()
+        _data_url = data_url or get_data_url()
+
+        self.trading_ws = _StreamConn(_key_id, _secret_key, _base_url)
+        self.data_ws = _StreamConn(_key_id, _secret_key, _data_url)
+        self.polygon = polygon.StreamConn(_key_id +
+            '-staing' if 'staging' in _base_url else '')
```
tibkiss, line 147: typo: `-staing`

PR #182, opened 2020-04-22
The Polygon Quote mapping did not change

shlomikushchi:
we need to use the same mapping as before

### DIFF
--- a/alpaca_trade_api/polygon/entity.py
+++ b/alpaca_trade_api/polygon/entity.py
@@ -329,15 +329,15 @@ trade_mapping = {
     "t": "timestamp"
 }
 
-quote_mapping = {
-    "T": "symbol",
-    "X": "askexchange",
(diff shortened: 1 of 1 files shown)


umitanuki:
@gnvk

gnvk:
@shlomikushchi my bad, the polygon mapping didn't change indeed, but the new endpoint's mapping is different. This PR will break that.  We need to use different mappings for the two.

shlomikushchi:
@gnvk ok, let me know when you get the right mapping. 
right now the backtrader integration is not working with 0.47rc2 so I hope this will help resolve that too

> @shlomikushchi my bad, the polygon mapping didn't change indeed, but the new endpoint's mapping is different. This PR will break that. We need to use different mappings for the two.

gnvk:
I'll merge this and then fix the new endpoint in a separate PR.

PR #184, opened 2020-04-23
Fix the quote mapping for the data steam

gnvk:


gnvk:
Closing this PR and moving it to origin so that the CI can run.

PR #185, opened 2020-04-23
Fix the quote mapping for the data steam

gnvk:


umitanuki:
I am not sure what this does and what @shlomikushchi tries to fix

PR #189, opened 2020-04-28
Simplify stream

umitanuki:
Now the caller needs to choose polygon explicitly, with the new data
streaming coming up

### DIFF
--- a/alpaca_trade_api/stream2.py
+++ b/alpaca_trade_api/stream2.py
@@ -87,6 +87,8 @@ class _StreamConn(object):
             raise ConnectionError("Max Retries Exceeded")
 
     async def subscribe(self, channels):
+        if isinstance(channels, str):
+            channels = [channels]
         if len(channels) > 0:
             await self._ensure_ws()
             self._streams |= set(channels)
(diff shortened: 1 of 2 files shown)


alpaca_trade_api/stream2.py
```diff
@@ -158,15 +166,31 @@ def __init__(
...
-            _key_id + '-staging' if 'staging' in _base_url else _key_id)
+
+        if self._data_stream == 'polygon':
+            self.data_ws = polygon.StreamConn(
+                _key_id + '-staging' if 'staging' in _base_url else _key_id)
+            self._data_prefixes = (('Q.', 'T.', 'A.', 'AM.'))
+        else:
+            self.data_ws = _StreamConn(_key_id, _secret_key, _data_url)
+            self._data_prefixes = (
+                ('Q.', 'T.', 'AM.', 'polyfeed/', 'alpacadatav1/'))
```
shlomikushchi, line 193: why both(polyfeed and alpacadatav1)? it's RC.. you don't need to keep supporting it.
also - now that it's an argument we don't need the prefix anymore. just use the same as polygon.

umitanuki: sure... thought just to keep the compatiblity for short. will remove it.

PR #191, opened 2020-05-01
Fix exception handling, timestamp precisions

umitanuki:


alpaca_trade_api/entity.py
```diff
@@ -134,6 +134,8 @@ def __getattr__(self, key):
         if key in self._raw:
             val = self._raw[key]
             if key == 'timestamp':
+                if val > 1000000000000000000:
```
shlomikushchi, line 137: you should document why we do this.
what is `1000000000000000000`
and it should be a const, not a raw number

umitanuki: sure. the reason is to accept nanoseconds correctly and to avoid OverflowError. I will add comments.

PR #194, opened 2020-05-04
create new event_loop for threads

samuel-soubeyran:
Backtrader library run strategies in threads. Those don't have an event loop, causing a runtime exception.
See: https://forum.alpaca.markets/t/alpaca-backtrader-api-gives-threading-error-on-running-sample-code/528


umitanuki:
The assumption is actually for the caller to explicitly create loop if they know they are running in threads. This exception handling is not to catch that case I believe...

PR #196, opened 2020-05-05
Not all errors are authentication errors. 

shlomikushchi:
handling also general exception


umitanuki:
What is pending here?

PR #202, opened 2020-05-19
Martingale Example - Support for the new alpaca data API

shlomikushchi:
udpating this example code to be able to choose which data source to use. 
originl examle used polygon second aggs
now we also enabled to use alpaca quote data 
to make it similar we don't act on every quote, unless 1 second past.

to change between data sources, change the global USE_POLYGON

### DIFF
--- a/examples/martingale.py
+++ b/examples/martingale.py
@@ -96,47 +96,14 @@ class MartingaleTrader(object):
                 tick_close = data.close
                 self.last_price = tick_close
 
-                # Update streak info
-                diff = truncate(tick_close, 2) - truncate(tick_open, 2)
-                if diff != 0:
-                    # There was a meaningful change in the price
-                    self.streak_count += 1
-                    increasing = tick_open > tick_close
-                    if self.streak_increasing != increasing:
-                        # It moved in the opposite direction of the streak.
-                        # Therefore, the streak is over, and we should reset.
-
-                        # Empty out the position
-                        self.send_order(0)
-
-                        # Reset variables
-                        self.streak_increasing = increasing
-                        self.streak_start = tick_open
-                        self.streak_count = 0
-                    else:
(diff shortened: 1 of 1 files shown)


examples/martingale.py
```diff
@@ -84,40 +96,27 @@ async def handle_agg(conn, channel, data):
...
+        # Listen for quote data and perform trading logic
+        @conn.on(r'T\..+', [self.symbol])
+        async def handle_alpaca_aggs(conn, channel, data):
+            now = datetime.datetime.utcnow()
+            if now - self.last_trade_time < datetime.timedelta(seconds=1):
+                # don't react every tick unless at least 1 second past
+                return
+            self.last_trade_time = now
+            self.tick_index = (self.tick_index + 1) % (self.tick_size)
+            print(self.tick_index)
```
umitanuki, line 110: Do you intend to keep this print? If so, I would suggest putting more readable message.

examples/martingale.py
```diff
@@ -146,7 +145,45 @@ async def handle_trade(conn, channel, data):
             elif event_type != 'new':
                 print(f'Unexpected order event type {event_type} received')
 
-        conn.run([f'A.{self.symbol}', 'trade_updates'])
+        if USE_POLYGON:
+            conn.run([f'A.{self.symbol}', 'trade_updates'])
+        else:
+            conn.run([f'alpacadatav1/T.{self.symbol}', 'trade_updates'])
+
+    async def process_current_tick(self, tick_open, tick_close):
```
umitanuki, line 153: Why is this async?

examples/martingale.py
shlomikushchi: deleted the print

examples/martingale.py
shlomikushchi: fixed

PR #203, opened 2020-05-20
Add Python typings hints to the SDK

shlomikushchi:
adding type hints will help the users to work the sdk more easily and therefore more effectivly with the API.

### DIFF
--- a/alpaca_trade_api/polygon/rest.py
+++ b/alpaca_trade_api/polygon/rest.py
@@ -85,7 +85,6 @@ class REST(object):
                            limit: int = None
                            ) -> TradesV2:
         """
-        https://
         polygon.io/docs/#get_v2_ticks_stocks_trades__ticker___date__anchor
         :param symbol
         :param date: DATE in this format YYYY-MM-DD
@@ -133,7 +132,6 @@ class REST(object):
                            limit: int = None
                            ) -> QuotesV2:
         """
-        https://
         polygon.io/docs/#get_v2_ticks_stocks_nbbo__ticker___date__anchor
         :param symbol
         :param date: DATE in this format YYYY-MM-DD
--- a/alpaca_trade_api/polygon/rest.py
+++ b/alpaca_trade_api/polygon/rest.py
@@ -1,4 +1,5 @@
-import dateutil
+from typing import List
+import dateutil.parser
 import requests
 from .entity import (
     Aggsv2, Aggsv2Set,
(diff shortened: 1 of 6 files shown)


alpaca_trade_api/polygon/rest.py
```diff
@@ -59,10 +76,25 @@ def historic_trades(self, symbol, date, offset=None, limit=None):
...
+    def historic_trades_v2(self,
+                           symbol: str,
+                           date: DATE,
+                           timestamp: int = None,
+                           timestamp_limit: int = None,
+                           reverse: bool = None,
+                           limit: int = None
+                           ) -> TradesV2:
+        """
+        https://
```
umitanuki, line 88: Do you want to break line here?

alpaca_trade_api/rest.py
```diff
@@ -220,16 +243,41 @@ def list_orders(self, status=None, limit=None, after=None, until=None,
...
+                     order_class: str = None,
+                     take_profit: dict = None,
+                     stop_loss: dict = None):
+        """
+        :param symbol: symbol or asset ID
+        :param qty: int
+        :param side: buy or sell
+        :param type: market, limit, stop, or stop_limit
+        :param time_in_force: day, gtc, opg, cls, ioc, fok
+        :param limit_price: str of float
```
umitanuki, line 265: I believe lots of users put some number type here. Does it break such code or not?

umitanuki: In this cas, str is preferred but nothing prevents int/float. What would be the right hinting?

alpaca_trade_api/rest.py
```diff
@@ -220,16 +243,41 @@ def list_orders(self, status=None, limit=None, after=None, until=None,
...
+        :param stop_price: str of float
+        :param client_order_id:
+        :param extended_hours:
```
umitanuki, line 268: Missing description?

umitanuki:
Please make sure backward compatibility with this change including those ambiguous cases like str/float parameters. Please let me know if you have any concerns.

alpaca_trade_api/polygon/rest.py
shlomikushchi: if I don't the line is too long...

alpaca_trade_api/rest.py
shlomikushchi: general guidelines should always be to give one way of operation. all users call the apis in the same way. it prevents a lot of headache in example code, documentation, errors and different behavior in different use cases(in my opinion) 
so the hints should be as such.
having said that - typings are not enforced, a user could decide for instance to pass a float here. the interpreter will allow it. 
the typings are used to help all users write the code in the same way (a guideline)

alpaca_trade_api/polygon/rest.py
shlomikushchi: fixed

alpaca_trade_api/rest.py
shlomikushchi: fixed

shlomikushchi:
> 
> 
> Please make sure backward compatibility with this change including those ambiguous cases like str/float parameters. Please let me know if you have any concerns.

both these API calls resulted in successful orders:
```py
api.submit_order("AAPL", 1, side='buy', type='limit',time_in_force='day',limit_price=331.6)
api.submit_order("AAPL", 1, side='buy', type='limit',time_in_force='day',limit_price='331.6')
```

as I mentioned typings are not enforced by the interpreter so it shouldn't break anything

PR #204, opened 2020-05-23
If a websocket disconnects - try to reconnect and subscribe

shlomikushchi:
if we get a disconnected ws meaning consume() is completed 
we close the connection for both data ws and trade ws
now we added a check - should we reconnect on close, and we re open the connections and subscribe to the same channels.

### DIFF
--- a/alpaca_trade_api/stream2.py
+++ b/alpaca_trade_api/stream2.py
@@ -288,7 +288,7 @@ class StreamConn(object):
                 logging.info("Exiting on Interrupt")
                 should_renew = False
             except Exception as e:
-                print(e)
+                logging.error(f"error while consuming ws messages: {e}")
                 loop.run_until_complete(self.close(should_renew))
                 if loop.is_running():
                     loop.close()
--- a/alpaca_trade_api/stream2.py
+++ b/alpaca_trade_api/stream2.py
@@ -181,9 +181,9 @@ class StreamConn(object):
             base_url=None,
             data_url=None,
             data_stream=None):
-        _key_id, _secret_key, _ = get_credentials(key_id, secret_key)
-        _base_url = base_url or get_base_url()
-        _data_url = data_url or get_data_url()
+        self._key_id, self._secret_key, _ = get_credentials(key_id, secret_key)
+        self._base_url = base_url or get_base_url()
+        self._data_url = data_url or get_data_url()
         if data_stream is not None:
             if data_stream in ('alpacadatav1', 'polygon'):
                 _data_stream = data_stream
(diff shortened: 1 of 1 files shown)


alpaca_trade_api/stream2.py
```diff
@@ -271,23 +276,46 @@ def run(self, initial_channels=[]):
...
+                if loop.is_closed():
+                    self.loop = asyncio.new_event_loop()
+                    loop = self.loop
+                loop.run_until_complete(self.subscribe(initial_channels))
+                loop.run_until_complete(self.consume())
+            except KeyboardInterrupt:
+                logging.info("Exiting on Interrupt")
+                should_renew = False
+            except Exception as e:
+                print(e)
```
umitanuki, line 291: Do you want to use `logger.error()`

alpaca_trade_api/stream2.py
```diff
@@ -271,23 +276,46 @@ def run(self, initial_channels=[]):
...
+        while should_renew:
+            try:
+                if loop.is_closed():
+                    self.loop = asyncio.new_event_loop()
+                    loop = self.loop
+                loop.run_until_complete(self.subscribe(initial_channels))
+                loop.run_until_complete(self.consume())
+            except KeyboardInterrupt:
+                logging.info("Exiting on Interrupt")
+                should_renew = False
```
umitanuki, line 289: How effective is this? It won't use this variable in the case of other exception?

umitanuki: Oh I see, the it exists the loop.

shlomikushchi: basically we want to reconnect for any exception not handled by the user

shlomikushchi: I would even remove the "except KeyboardInterrupt" because if the user breaks execution everything stops anyway

alpaca_trade_api/stream2.py
shlomikushchi: fixed

alpaca_trade_api/stream2.py
```diff
@@ -271,23 +276,46 @@ def run(self, initial_channels=[]):
...
+        Close any of open connections
+        :param renew: should re-open connection?
+        """
         if self.trading_ws is not None:
             await self.trading_ws.close()
             self.trading_ws = None
         if self.data_ws is not None:
             await self.data_ws.close()
             self.data_ws = None
+        if renew:
```
jpwinans, line 307: When this renews, it doesn't seem to retain the channels previously handled. After this renewal, the consumer just hangs. Is there an example of using this stream connection with this new code?

Issue #205, opened 2020-05-24
historic_agg_v2 TypeError: Parser must be a string or character stream, not Timestamp


Devination:
You used to be able to pass datetime objects to the v1 version of this function. Now it fails with these error messages.
```
TypeError: Parser must be a string or character stream, not Timestamp
TypeError: Parser must be a string or character stream, not datetime
```

shlomikushchi:
I added some comments on your PR (required fixes to make it a bit better)
but
I will copy the same note I wrote there too:
in my opinion - we should select one acceptable format for all users.
if we do accept this change - users could use 4 different formats.
there's no need for such flexability (it will cause confusion and contradicting examples/documentation in the long run)

if we do allow one argument type only, the users will do the conversions on their end (and it's a very simple conversion) and we will have consistancy.

shlomikushchi:
merged in https://github.com/alpacahq/alpaca-trade-api-python/pull/210

PR #206, opened 2020-05-24
Fix historic_agg_v2 TypeError issue #205

Devination:
Please let me know if there are any changes you would like me to make.

### DIFF
--- a/alpaca_trade_api/polygon/rest.py
+++ b/alpaca_trade_api/polygon/rest.py
@@ -1,3 +1,4 @@
+import datetime
 import dateutil
 import requests
 from .entity import (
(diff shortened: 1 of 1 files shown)


alpaca_trade_api/polygon/rest.py
```diff
@@ -134,21 +135,19 @@ def historic_agg_v2(self, symbol, multiplier, timespan, _from, to,
         path_template = '/aggs/ticker/{symbol}/range/{multiplier}/' \
                         '{timespan}/{_from}/{to}'
         if isinstance(_from, int):
-            path = path_template.format(symbol=symbol,
+            pass
```
shlomikushchi, line 138: so, if we don't do anything if it's an int we don't need the if

alpaca_trade_api/polygon/rest.py
```diff
@@ -134,21 +135,19 @@ def historic_agg_v2(self, symbol, multiplier, timespan, _from, to,
...
+        elif isinstance(_from, datetime.datetime):
```
shlomikushchi, line 139: if we do take care of that case, we should also check for `datetime.date`

alpaca_trade_api/polygon/rest.py
```diff
@@ -134,21 +135,19 @@ def historic_agg_v2(self, symbol, multiplier, timespan, _from, to,
...
+            _from = _from.date().isoformat()
+            to = to.date().isoformat()
```
shlomikushchi, line 141: this entire code segment should be in a method called: format_date()
which will make sure _from and to are in the right format.

Devination:
Your suggestions are good, especially the format_date() method which I'm sure would be useful in other parts of the API I haven't looked at yet.

As far as selecting only 1 date format for the API, that's a design decision for Alpaca. Feel free to discard my pull request if that's what Alpaca wants. Personally I'd prefer to add the flexibility in the API to make it as easy as possible to work with and avoid repeated conversion code in all users' projects.

It's possible that allowing multiple formats might cause confusion but I think people would understand fairly quickly that the API is flexible once they see 2 different examples of different date formats being used.

umitanuki:
I would say it's probably common to accept multiple data types especially this comes with pandas background. I found it useful to have this flexibility myself. We need comprehensive documentation to make it clear what is accepted and how they are converted, both in docstring and README. Can you make that change here before merging?

alpaca_trade_api/polygon/rest.py
umitanuki: Also, I would handle each parameter separately instead of assuming both are in the same type.

shlomikushchi:
was merged to https://github.com/alpacahq/alpaca-trade-api-python/tree/historic_agg_v2_date_type to make the fixes required locally and then merge to master with: https://github.com/alpacahq/alpaca-trade-api-python/pull/210

Issue #213, opened 2020-06-11
polygon/REST.daily_open_close(symbol, date) uses Python keyword 'from'


duhfrazee:
The DailyOpenClose entity uses 'from' for the date field being returned. This is a python keyword and causes a syntax error when trying to access it. See examples below:

```python
data = live_api.polygon.daily_open_close('AAPL', '2020-05-29')

print(data)

# DailyOpenClose({   'afterHours': 319,
#    'close': 317.94,
#    'from': '2020-05-29',
#    'high': 321.15,
#    'low': 316.47,
#    'open': 319.25,
#    'preMarket': 318.09,
#    'status': 'OK',
#    'symbol': 'AAPL',
#    'volume': 38670679})

print(data.close)
# 317.94

print(data.from)
# File "<ipython-input-66-b74cad23120f>", line 1
#     data.from
#             ^
# SyntaxError: invalid syntax
```

jaggas:
I recommend using data['from'] for element access, it is the preferred way in python.

duhfrazee:
Unfortunately data is not scriptable and you will receive the following error using that syntax

```python
print(data['from'])

# TypeError                                 Traceback (most recent call last)
# <ipython-input-75-0993a8d7e0c8> in <module>
# ----> 1 data['from']
# 
# TypeError: 'DailyOpenClose' object is not subscriptable

```

And thank you @shlomikushchi for the insane response time!

PR #216, opened 2020-06-11
Can't use 'from' as an attribute in python

shlomikushchi:
a possible fix for this issue: https://github.com/alpacahq/alpaca-trade-api-python/issues/213
please review and let me know if you like this solution

### DIFF
--- a/alpaca_trade_api/polygon/entity.py
+++ b/alpaca_trade_api/polygon/entity.py
@@ -11,7 +11,6 @@ class Entity(object):
             # can't use python keyword 'from'. if the api returns it,
             # we switch it to _from, which is usable for the users.
             self._raw['_from'] = self._raw['from']
-            del self._raw['from']
 
     def __getattr__(self, key):
         if key in self._raw:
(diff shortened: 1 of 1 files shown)


alpaca_trade_api/polygon/entity.py
```diff
@@ -7,6 +7,11 @@
 class Entity(object):
     def __init__(self, raw):
         self._raw = raw
+        if 'from' in self._raw:
+            # can't use python keyword 'from'. if the api returns it,
+            # we switch it to _from, which is usable for the users.
+            self._raw['_from'] = self._raw['from']
+            del self._raw['from']
```
umitanuki, line 14: you don't need to delete it do you?

shlomikushchi: ok I will not delete it.
doing it on getattr will not work since the error is invoked by the python ide
you can't write something like this:
```py
data = live_api.polygon.daily_open_close('AAPL', '2020-05-29')
data.from
```

PR #219, opened 2020-06-13
Pyup integration

shlomikushchi:
restructure the build structure a bit.
instead of specifying the requirements in the setup.py we now have a requirements folder for prod and test.
pyup can't scan setup.py, it requires files (requirements.txt files)

also adding the badges to the README file


requirements/requirements.txt
```diff
@@ -0,0 +1,8 @@
+asyncio-nats-client
```
umitanuki, line 1: We actually don't need this anymore.

shlomikushchi: removing it

PR #235, opened 2020-06-23
validate stop/limit prices before passed to the server

shlomikushchi:
we could pass stop/limit prices as floats or as string float to the api
when you add this flexibility, you may cause errors by wrong usage
e.g we could pass "a" and the request will go to the server but the result will be bad
by adding this, we will raise an exception before the request leaves the user's code
allowing to find errors earlier at runtime

### DIFF
--- a/alpaca_trade_api/common.py
+++ b/alpaca_trade_api/common.py
@@ -41,6 +41,26 @@ class DATE(str):
         return str.__new__(cls, value)
 
 
+class FLOAT(str):
+    """
+    api allows passing floats or float as strings.
+    let's make sure that param passed is one of the two, so we don't pass
+    invalid strings all the way to the servers.
+    """
+    def __new__(cls, value):
+        if not value:
+            raise ValueError('Unexpected empty string')
+        if isinstance(value, float) or isinstance(value, int):
+            pass  # we're good
+        elif isinstance(value, str):
+            value = value.strip()  # make sure no spaces
+            if isinstance(float(value), float):
+                pass  # we're good
+        else:
+            raise ValueError(f'Unexpected float format "{value}"')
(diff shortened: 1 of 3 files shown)


alpaca_trade_api/common.py
```diff
@@ -41,6 +41,26 @@ def __new__(cls, value):
...
+    """
+    def __new__(cls, value):
+        if not value:
+            raise ValueError('Unexpected empty string')
+        if isinstance(value, float) or isinstance(value, int):
+            pass  # we're good
+        elif isinstance(value, str):
+            value = value.strip()  # make sure no spaces
+            if isinstance(float(value), float):
+                pass  # we're good
```
umitanuki, line 58: if not? or if you are expecting exception what's the point of the `if`?

alpaca_trade_api/rest.py
```diff
@@ -282,9 +282,9 @@ def submit_order(self,
             'time_in_force': time_in_force,
         }
         if limit_price is not None:
-            params['limit_price'] = limit_price
+            params['limit_price'] = FLOAT(limit_price)
         if stop_price is not None:
-            params['stop_price'] = stop_price
+            params['stop_price'] = FLOAT(stop_price)
```
umitanuki, line 287: Do you want to keep it consistent in the bracket legs as well?

alpaca_trade_api/common.py
shlomikushchi: I guess it's good for readability. reader know we've thought about this scenario

alpaca_trade_api/rest.py
shlomikushchi: done

PR #245, opened 2020-06-28
Require pandas<0.23

shlomikushchi:
pylivetrader is using zipline 1.3.0
that package has that constraint.
we add it here too


shlomikushchi:
Already done in a different PR

Issue #252, opened 2020-07-02
Loosen up the requirements


mLewisLogic:
Related to: https://github.com/alpacahq/alpaca-trade-api-python/issues/230

Specifying exact versions for a 3rd party package is not a very friendly practice.
Locking to exact versions is critical for a production application, but for libraries that will be consumed by others, you generally want version specifications as wide as possible (without breakage).

Python has its warts with dependency management, but locking to specific versions is going to magnify those by an order of magnitude and virtually guarantee version incompatibilities as an app grows (along with its dependencies).

mLewisLogic:
For color, this doubly goes for "basic" libs (like requests, urllib, ws, etc) that basically every sizeable networking app is bound to inherit as a dep sooner or later.

I'm looking specifically at these:

# pandas<0.23
What breaks if `alpaca-trade-api-python` is run with Pandas `0.23`? This holds a Pandas-dependent app back from upgrading. I'm sure you can imagine that your users would like to do data analysis as part of their trading operations.


# requests==2.24.0
Way too specific here.

# urllib3==1.24.3
Way too specific as well

# websocket-client==0.57.0
Same.

# alpha_vantage==2.2.0
This one is probably fine given that it's an esoteric lib that is less likely of hitting conflict. Again though, why the exact specification here? Maybe it needs [asyncio support](https://github.com/RomelTorres/alpha_vantage#news)? Probably makes sense to switch it to a `>=2.2.0` in that case to unlock minor version upgrades.

mLewisLogic:
Actually, was thinking about the Pandas issue a little more. That really should be left open-ended in this lib.
If `pylivetrader` wants to specify a specific Pandas requirement (due to `zipline`), that can be done there. There's no reason to push that requirement down into this repo.

mLewisLogic:
Related to: https://github.com/alpacahq/alpaca-trade-api-python/issues/242

shlomikushchi:
Hi
thanks for the comments
about pandas, basically we wanted the behavior to be the same all across our packages.
but when using pylivetrader, the user doesn't have direct access to the alpaca-trade-api, so I'm thinking it's ok to have different restrictions (I am a bit concerned about different behavior underneath the interface, but it should probably be ok)

about the other packages, we use pyup, which requires explicit package versioning. the logic is that we always upgrade to the latest version and that this should not be an issue. we just started the integration so it will take some time to really work smoothly.

mLewisLogic:
[...]

PR #262, opened 2020-07-07
Loosen the requirements to easily install with other packages

shlomikushchi:
as a response to this issue: https://github.com/alpacahq/alpaca-trade-api-python/issues/252

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,6 +1,6 @@
-pandas<0.23
-requests==2.24.0
-urllib3==1.24.3
-websocket-client==0.57.0
-websockets>=8.0
-alpha_vantage==2.2.0
+pandas>0.22
+requests>=2.24.0,<3
+urllib3>=1.24.3,<1.26
+websocket-client>=0.56.0,<1
+websockets>=8.0,<9
+alpha_vantage>=2,<3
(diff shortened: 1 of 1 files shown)


requirements/requirements.txt
```diff
@@ -1,6 +1,6 @@
-pandas<0.23
-requests==2.24.0
-urllib3==1.24.3
-websocket-client==0.57.0
-websockets>=8.0
-alpha_vantage==2.2.0
+pandas>=0.18.1
```
umitanuki, line 1: I would agree more to loosen other packages. Do we want to support both pandas 0.x and 1.x?

shlomikushchi: we could with:
pandas 
or
pandas>1

and let pylivetrader downgrade it

umitanuki: I'm inclined to do >1 then.

umitanuki: But does it actually work? I think pip will reject such install?

umitanuki: If pylivetrader can't support pandas>1, then maybe we need to do >=0.xx

shlomikushchi: pip will give a warning but will install correctly (for pylivetrader it will remove 1.0.5 and install 0.22.0)

PR #283, opened 2020-07-27
Make sure that historic_aggs returns the 'to' data for daily bars

shlomikushchi:
the polygon api does not include the end date for daily bars, or this:
```
    historic_agg_v2("SPY", 1, "day", _from="2020-07-22", to="2020-07-24").df
    results in this:
    timestamp
    2020-07-22 00:00:00-04:00  324.62  327.20  ...  57917101.0  325.8703
    2020-07-23 00:00:00-04:00  326.47  327.23  ...  75841843.0  324.3429
```
    the 24th data is missing
    for minute bars, it does include the end date

    so basically this method will add 1 day (if 'to' is not today, we don't
    have today's data until tomorrow) to the 'to' field

### DIFF
--- a/alpaca_trade_api/polygon/rest.py
+++ b/alpaca_trade_api/polygon/rest.py
@@ -46,6 +46,30 @@ def format_date_for_api_call(date):
         raise Exception(f"Unsupported date format: {date}")
 
 
+def fix_daily_bar_date(date, timespan):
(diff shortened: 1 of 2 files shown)


shlomikushchi:
I've only seen it in `historic_agg_v2` thus far

PR #285, opened 2020-07-27
Increase ws consume() exceptions verbosity

shlomikushchi:
When we get ws exceptions in the consume() method, it's usually not
informative enough. but, we may only want to see more information when we debug
by adding the debug flag to the StreamConn object we will add a stacktrace when
consume() related exceptions occur, and give the user more information helping them
to the debug the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,19 @@ same `channel_pat` will overwrite the old handler.
 Deregisters the event handler function that was previously registered via `on` or
 `register` method.
 
+#### Debugging
+Websocket exceptions may occur during execution.
+It will usually happen during the `consume()` method, which basically is the 
+websocket steady-state.<br>
+exceptions during the consume method may occur due to:
+- server disconnections
+- error while handling the response data
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -285,6 +285,19 @@ same `channel_pat` will overwrite the old handler.
...
+websocket steady-state.<br>
+exceptions during the consume method may occur due to:
+- server disconnections
+- error while handling the response data
+
+We handle the first issue by reconnecting the websocket every time there's a disconnection.
+The second issue, is usually a user's code issue. To help you find it, we added a flag to the 
+StreamConn object called `debug`. It is set to False by default, but you can turn it on to get a more
+verbose logs when this exception happens.
+Turn it on like so `StreamConn(debug=True)`
```
umitanuki, line 300: Does it make sense to add the flag to REST as well?

shlomikushchi: I will eventually. this is an error that happens a lot inside the loop, and the REST module doesn't have a loop.
the errors of the REST module are more straight forward. when I'll find errors that aren't I will add this too.

PR #304, opened 2020-08-28
Fix hanging close after Alpaca WebSocketException, add closing_timeout

drew-wallace:
This fixed an issue I was having where the Alpaca websocket would throw a warning and hang forever. Turns out, it was awaiting on closing the socket even though it was already closed. Setting the socket variable to None and letting it continue to _ensure_ws fixed the issue.

I also added a closing_timeout to the socket connection so it closes the fastest possible.

Finally, I fixed the streams variable being re-passed into subscribe as a Set instead of a List.


nickvertucci:
There are 2 PRs with changing to a list on reconnect. LETS GET THIS DONE

Issue #109, opened 2019-08-18
get_barset() still ignores start and end date


edwardyu:
I've looked through #66 and the advice doesn't work.

```
bars = api.get_barset(['AAPL'], '1D', limit=100, start=pd.Timestamp('2010-01-01', tz='America/New_York').isoformat())
print(bars.df)
```

```
                               AAPL                                        
                               open      high       low     close    volume
time                                                                       
2019-03-27 00:00:00-04:00  188.7500  189.7600  186.5500  188.4700  26586290
2019-03-28 00:00:00-04:00  188.9500  189.5590  187.5300  188.7100  19309376
2019-03-29 00:00:00-04:00  189.9900  190.0800  188.5400  189.9400  18489986
2019-04-01 00:00:00-04:00  191.6400  191.6800  188.3800  191.2400  12098014
2019-04-02 00:00:00-04:00  191.0900  194.4600  191.0500  194.0400  19797454
2019-04-03 00:00:00-04:00  193.2500  196.5000  193.1500  195.3500  19793141
2019-04-04 00:00:00-04:00  194.7900  196.3700  193.1400  195.7200  17091108
2019-04-05 00:00:00-04:00  196.4500  197.1000  195.9300  196.9700  15842943
2019-04-08 00:00:00-04:00  196.4200  200.2300  196.3400  200.0800  23231403
2019-04-09 00:00:00-04:00  200.3200  202.8500  199.2300  199.5000  32840959
2019-04-10 00:00:00-04:00  198.6800  200.7400  198.1800  200.6600  19861942
2019-04-11 00:00:00-04:00  200.8500  201.0000  198.4431  198.9500  17931789
2019-04-12 00:00:00-04:00  199.2000  200.1400  196.2100  198.8800  24667735
2019-04-15 00:00:00-04:00  198.5800  199.8500  198.0100  199.2300  14551176
2019-04-16 00:00:00-04:00  199.4600  201.3700  198.5600  199.2500  24002891
2019-04-17 00:00:00-04:00  199.5400  203.3800  198.6100  203.1200  27366117
2019-04-18 00:00:00-04:00  203.1200  204.1500  202.5200  203.8600  21928367
2019-04-22 00:00:00-04:00  202.8300  204.9400  202.3400  204.6400  13720923
2019-04-23 00:00:00-04:00  204.4300  207.7500  203.9000  207.5100  19401417
[...]

PR #309, opened 2020-09-09
Example using get_barset

shlomikushchi:
exactly how to use this api

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,20 +36,6 @@ account = api.get_account()
 api.list_positions()
 ```
 
-#### Using `get_barset()`
-```python 
-import pandas as pd
-NY = 'America/New_York'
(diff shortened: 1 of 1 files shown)


keatbro:
@shlomikushchi  I tried your example from [here](https://github.com/alpacahq/alpaca-trade-api-python/issues/109#issuecomment-689503055), but if I change either of the year inputs (on either the start or end date), it doesn't run.

For example, instead of your example of a start date of:
`start=pd.Timestamp('2020-08-01', tz=NY).isoformat()`

I simply change the year, to:
`start=pd.Timestamp('2019-08-01', tz=NY).isoformat()`

And suddenly it only returns 100 rows (indicating it reach the default **limit** value, I think). Any ideas?

shlomikushchi:
probably means you need to iterate on smaller time frames

PR #314, opened 2020-09-22
fixed typo in simpleStream.py

eric-berglund:


eric-berglund (commented):
First time contributing to a project like this on GitHub. Finding it really interesting and have just been fixing minor typos here and there. Let me know if this is helpful or if I am going about this the right way or not.

PR #326, opened 2020-11-24
Add note when using get_barset with limit

shlomikushchi:
When using get_barset() with limit, the count is from the end date
if end date is not specified, "now" is being used.
that could be confusing when using start date with limit and start date is something like (2016)
the user expects the data to be in 2016, if "now" is used, it will be only 2020 so the user need to be aware of that.


umitanuki (approved):
The doc will help for now, but I'm open to ideas for better API design.

PR #327, opened 2020-11-27
Make AggsV2 more robust for None returned values

medusa-trade:
The previous implementation can error if the return value from Polygon.io is empty and `self._raw` is `None`. This implementation is more robust.

### DIFF
--- a/alpaca_trade_api/polygon/entity.py
+++ b/alpaca_trade_api/polygon/entity.py
@@ -1,6 +1,8 @@
 import pandas as pd
 import pprint
 
+from collections import defaultdict
+
 NY = 'America/New_York'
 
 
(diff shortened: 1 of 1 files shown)


shlomikushchi:
Hi,
did you get this response from the api? when I tested that I got an empty list for `_raw`, not `None`

medusa-trade:
@shlomikushchi - Yes, using what's in `master` I received a `None` value for `_raw`. I think this is very intermittent from Polygon, because it was fine after rerunning it. I think my solution will work for both cases.

alpaca_trade_api/polygon/entity.py
```diff
@@ -93,14 +95,8 @@ def __init__(self, raw):
...
     def _raw_results(self):
-        results = self._raw.get('results')
-        if not results:
-            # this is not very pythonic but it's written like this because
-            # the raw response for empty aggs was None, and this:
-            # self._raw.get('results', []) returns None, not [] which breaks
-            # when we try to iterate it.
-            return []
-        return results
+        self._raw = self._raw or defaultdict(lambda: [])
```
shlomikushchi, line 98: sorry for the late response (your mixed closed PR with pickle data brought this back to my attention)
didn't merge this back then because this is not answering the issue I wrote the initial code for.
the scenario in which there is a response but there is no `results` in the response. 

this code, in that case will result in an exception. an example: trying to do something like this: 
`api.polygon.historic_agg_v2('AAPL2', 1, 'day', _from='2019-01-01', to='2019-1-5').df`
note that is a request for a non existing symbol: `AAPL2`

this will be the result of this code:
image

shlomikushchi:
I will use this PR that handles both cases: https://github.com/alpacahq/alpaca-trade-api-python/pull/346

shlomikushchi:
merged #346

PR #334, opened 2020-12-08
Fallback to use environment version

johnslay:
Setting v2 on the api or through an environment version is overridden by this when calling any api method.


shlomikushchi:
Hi, `data_get` is used by 4 methods:
* get_last_trade
* get_last_quote
* get_barset
* get_aggs (old. undocumented so not relevant to this issue.)

all 3 relevant methods use alpaca api V1:
image

image

image

this is why v1 is used in this method explicitly.

PR #344, opened 2020-12-28
chore(trades): added fields to trade mapping

Jrenk:
I added `f`, `y` and `e` to trades. Without this, these values would get lost in this mapping.

### DIFF
--- a/alpaca_trade_api/polygon/entity.py
+++ b/alpaca_trade_api/polygon/entity.py
@@ -345,8 +345,12 @@ trade_mapping = {
     "s": "size",
     "t": "timestamp",
     "f": "trf_timestamp",
-    "y": "exchange_timestamp",
-    "e": "trade_correction"
+    "y": "participant_timestamp",
+    "e": "trade_correction",
+    "i": "id",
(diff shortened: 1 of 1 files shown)


alpaca_trade_api/polygon/entity.py
```diff
@@ -343,7 +343,10 @@ class DailyOpenClose(Entity):
     "x": "exchange",
     "p": "price",
     "s": "size",
-    "t": "timestamp"
+    "t": "timestamp",
+    "f": "trf_timestamp",
+    "y": "exchange_timestamp",
```
shlomikushchi, line 348: we use `participant_timestamp` not `exchange_timestamp` so should be probably changed to that

shlomikushchi: there are some other fields we should consider map: i, r, q, z

Jrenk: I renamend `exchange_timestamp` to `participant_timestamp` and added `i`, `r`, `q` and `z` to mapping.

umitanuki (approved):

PR #345, opened 2021-01-02
Updating Polygon Entity to allow pickle.load to work

medusa-trade:
Currently, if you `pickle.dump` any Entity that's been initialized, and you try to `pickle.load` that object, you will receive an infinite regression. I've added following code so initialization on `pickle.load` works correctly. Not sure if this would be useful for anyone else who wants to save time with historical backtesting from data already loaded.


medusa-trade:
This incorrectly included code from a previous PR. Deleting and will recreate.

PR #351, opened 2021-01-05
"Set not serializable" error for Python 3.8.7 json.dumps

medusa-trade:
With Python 3.8.7, I receive `TypeError: Object of type set is not JSON serializable` exception when `_StreamConn` tries to `json.dump` the channels after `suscribe()` is called with `self._streams` in `_ensure_ws()` 

`list(channels)` should always be serializable.


medusa-trade:
Need to clean PR offline from Github UI.

PR #352, opened 2021-01-05
Always serialize channels and streams as list

medusa-trade:
On Python 3.8.7, I am receiving `TypeError: Object of type set is not JSON serializable` when a broken connection is trying to be reestablished.


medusa-trade:
Upon further testing, this solution doesn't work.

Shakespeare0:
@medusa-trade Why did the solution not work? I ran into the same problem and made a similar change #363 and it fixed the issue for me.

PR #353, opened 2021-01-05
get and set state in a way that makes Entity pickleable

medusa-trade:
This solution prevents an infinite regression when setting `_raw` during deserialization of a `pickle.dump` using `pickle.load` for a collection of `Entity` objects.

### DIFF
--- a/alpaca_trade_api/entity.py
+++ b/alpaca_trade_api/entity.py
@@ -34,6 +34,12 @@ class Entity(object):
             raw=pprint.pformat(self._raw, indent=4),
         )
 
+    def __getstate__(self):
+        return self._raw
+
+    def __setstate__(self, state):
+        self._raw = state
+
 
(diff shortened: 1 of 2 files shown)


shlomikushchi:
Hi, I am working on this PR right now: https://github.com/alpacahq/alpaca-trade-api-python/pull/355
It will be ready in a couple of days.
I think once it's ready there will be no need for this one.
Let me know your thoughts about it

medusa-trade:
I agree. With the ability to just use the `raw` JSON, this PR isn't necessary. It is nice, however, to use the Entity objects, and if the rest of my application is using the objects as `Entity` objects, it's easier to pickle them as the same object. But I don't think this is necessary if others won't find it useful. I submit to the will of the collective. :)

medusa-trade:
Feel free to close and delete this PR.

PR #363, opened 2021-01-13
Pass self._streams as a list to subscribe()

Shakespeare0:
If the connection drops and it trys to reconnect, it fails because the channels are saved in self._streams, which is a set, which is not serializable in JSON, so convert it back to a list when passing the channels to _StreamConn.subscribe().

### DIFF
--- a/alpaca_trade_api/polygon/streamconn.py
+++ b/alpaca_trade_api/polygon/streamconn.py
@@ -124,7 +124,7 @@ class StreamConn(object):
             try:
                 await self.connect()
                 if self._streams:
-                    await self.subscribe(self._streams)
+                    await self.subscribe(list(self._streams))
                 break
             except Exception as e:
                 await self._dispatch({'ev': 'status',
--- a/alpaca_trade_api/stream2.py
+++ b/alpaca_trade_api/stream2.py
@@ -99,7 +99,7 @@ class _StreamConn(object):
             try:
                 await self._connect()
                 if self._streams:
-                    await self.subscribe(self._streams)
+                    await self.subscribe(list(self._streams))
                 break
             except websockets.WebSocketException as wse:
                 logging.warn(wse)


shlomikushchi:
This PR was closed last week: https://github.com/alpacahq/alpaca-trade-api-python/pull/352
with the same solution but @medusa-trade came to the conclusion it doesn't work. could you elaborate on why?

medusa-trade:
@shlomikushchi / @Shakespeare0 - My implementation was a bit different. I was casting to `list` from within the json.dumps method within the `send`. In my case, it would just hang and not respond to messages. I didn't have time to fix, so I deleted the PR. By casting at the time of passing the arguments, it may work. Also, this bug seems to be in Python 3.8.7, and it works for me locally in Python 3.8.5. So this PR should be tested against both versions, IMO.

Shakespeare0:
I'm not sure why @medusa-trade 's fix did not work, but `self._streams` is [defined to be a set here](https://github.com/alpacahq/alpaca-trade-api-python/blob/f404cf14a35b594780da78e37bbab53cdb69d7bb/alpaca_trade_api/stream2.py#L34) and JSON does not support encoding sets, so when `self._streams` is used to reconnect if the connection drops [here](https://github.com/alpacahq/alpaca-trade-api-python/blob/f404cf14a35b594780da78e37bbab53cdb69d7bb/alpaca_trade_api/stream2.py#L108), it has to be converted back to a list at some point. In my opinion it should be before it is passed to `subscribe()`, since `subscribe()` tries to make sure it's argument is converted to a list (like [here](https://github.com/alpacahq/alpaca-trade-api-python/blob/f404cf14a35b594780da78e37bbab53cdb69d7bb/alpaca_trade_api/stream2.py#L120)) if it's not a list already.

medusa-trade:
I agree with @Shakespeare0. This seems like a better solution. I just haven't been able to test it. If it works across Python versions, I think it would be best to merge it. I honestly did dig into why my connections were hanging with my solution. I just know that when I removed those changes, it was still broke on my Ubuntu system with version 3.8.7, but it worked for my local 3.8.5. So I moved on.

Shakespeare0:
I think the connection hanging may be a different issue. Without this fix, rather than the connection hanging, I get the error `2021-01-15 22:27:28,344 error while consuming ws messages: Object of type set is not JSON serializable`.

nickvertucci:
I get 1006 errors ALLLLLLL THE TIME. Same thing Object of type set is not JSON serializable. Would love to see this fix merged in ASAP

nickvertucci:
Merge?

This is because Alpaca API gets disconnected midway. So the API tries to reconnect. But when it does so, the channels are not reinitialized, hence it remains a set that is not JSON serializable. Hence, in alpaca_trade_api > stream2.py change 'streams': channels to 'streams': list(channels) in line 132.
That should solve the reconnecting to api and 1006 problem.

PR #379, opened 2021-02-09
Advanced streamconn usage

shlomikushchi:
Add the ability to "pause/resume" the websocket connection.
also added advanced ws example code for:
* pause / resume connection
* change subscriptions of existing connection
* ws disconnections handler (make sure we reconnect)
* Different subscriptions usage


alpaca_trade_api/polygon/streamconn.py
```diff
@@ -190,12 +190,15 @@ def run(self, initial_channels=[]):
...
     async def close(self):
         '''Close any open connections'''
-        if self._consume_task:
-            self._consume_task.cancel()
+        await self.cancel_task()
         if self._ws is not None:
             await self._ws.close()
         self._ws = None
 
+    async def cancel_task(self):
```
gnvk, line 198: [minor] should be private: `_cancel_task()`

alpaca_trade_api/stream2.py
```diff
@@ -140,12 +141,15 @@ async def unsubscribe(self, channels):
...
 
     async def close(self):
-        if self._consume_task:
-            self._consume_task.cancel()
+        await self.cancel_task()
         if self._ws:
             await self._ws.close()
             self._ws = None
 
+    async def cancel_task(self):
```
gnvk, line 149: [minor] should be private: `_cancel_task()`

alpaca_trade_api/stream2.py
```diff
@@ -227,6 +231,7 @@ def __init__(
         self._data_stream = _data_stream
         self._debug = debug
         self._raw_data = raw_data
+        self.stop_stream_queue = queue.Queue()
```
gnvk, line 234: [minor] should be private: `_stop_stream_queue`

alpaca_trade_api/stream2.py
```diff
@@ -370,6 +378,19 @@ async def close(self, renew):
                                            self._oauth,
                                            raw_data=self._raw_data)
 
+    async def signal_stop_ws(self):
```
gnvk, line 381: [minor] I would simply call this `stop()` or `stop_ws()`

alpaca_trade_api/stream2.py
```diff
@@ -370,6 +378,19 @@ async def close(self, renew):
...
+        """
+        Signal the ws connections to stop.
+        We set a signal in a Queue and then call cancel_task
```
gnvk, line 384: [minor] this is an implementation detail that I wouldn't include in the docstring.

alpaca_trade_api/stream2.py
shlomikushchi: done

alpaca_trade_api/stream2.py
shlomikushchi: done

alpaca_trade_api/stream2.py
shlomikushchi: done. removed internal implementation from doc string

alpaca_trade_api/polygon/streamconn.py
shlomikushchi: it is used outside of the class so it shouldn't be private

alpaca_trade_api/stream2.py
shlomikushchi: same. used out side of the class.

gnvk (approved):
LGTM

PR #384, opened 2021-02-19
Data v2

gnvk:


doudoujay:
do you know when could we get V2 support with this client?

PR #391, opened 2021-03-04
Update v2_example.py to add paper credentials instruction comment

nickjohnson22:
This stumped me for a while. Hopefully, the comment helps others who want to test with their paper credentials.

### DIFF
--- a/examples/websockets/v2_example.py
+++ b/examples/websockets/v2_example.py
@@ -2,6 +2,9 @@ import logging
 
 from alpaca_trade_api.stream import Stream
 
+# Uncomment URL import to test with paper credentials
+# from alpaca_trade_api.common import URL
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


examples/websockets/v2_example.py
```diff
@@ -2,6 +2,9 @@
 
 from alpaca_trade_api.stream import Stream
 
+# Uncomment URL import to test with paper credentials
+# from alpaca_trade_api.common import URL
```
gnvk, line 6: No need for this import, you can simply use a string for `base_url`

examples/websockets/v2_example.py
```diff
@@ -20,7 +23,8 @@ async def print_trade_update(tu):
 def main():
     logging.basicConfig(level=logging.INFO)
     feed = 'iex'  # <- replace to SIP if you have PRO subscription
-    stream = Stream(data_feed=feed, raw_data=True)
+    
+    stream = Stream(data_feed=feed, raw_data=True)  # <- add base_url=URL('https://paper-api.alpaca.markets') for paper
```
gnvk, line 27: I would wrap it like this:
```suggestion
stream = Stream(
        data_feed=feed,
        # base_url='https://paper-api.alpaca.markets', # uncomment for paper
        raw_data=True)
```

PR #400, opened 2021-03-16
Update ws example code to data V2

camelpac:
Also upgrade Stream() to allow `stop_ws()`


alpaca_trade_api/stream.py
```diff
@@ -403,3 +430,15 @@ def run(self):
         except KeyboardInterrupt:
             print('keyboard interrupt, bye')
             pass
+
+    async def stop_ws(self):
+        """
+        Signal the ws connections to stop listenning to api stream.
+        """
+        if self._trading_ws:
+            logging.info("Stopping the trading websocket connection")
```
gnvk, line 439:
```suggestion
            log.info("Stopping the trading websocket connection")
```

alpaca_trade_api/stream.py
```diff
@@ -403,3 +430,15 @@ def run(self):
...
+            await self._trading_ws.stop_ws()
+
+        if self._data_ws:
+            logging.info("Stopping the data websocket connection")
```
gnvk, line 443:
```suggestion
            log.info("Stopping the data websocket connection")
```

gnvk (approved):
LGTM

PR #403, opened 2021-03-21
Added missing type hint py.typed files.

blane-nelson:
In order for utilities like MyPy to take advantage of the type hints that in the package there need to be py.typed files installed with the package as per [PEP 651](https://www.python.org/dev/peps/pep-0561/)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,10 @@ setup(
         'alpaca_trade_api',
         'alpaca_trade_api.polygon',
     ],
+    package_data={
+        'alpaca_trade_api': ['py.typed'],
+        'alpaca_trade_api.polygon': ['py.typed'],
+    },
     install_requires=REQUIREMENTS,
     tests_require=REQUIREMENTS_TEST,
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -35,6 +35,10 @@
         'alpaca_trade_api',
         'alpaca_trade_api.polygon',
     ],
+    package_data={
+        'alpaca_trade_api': ['py.typed'],
+        'alpaca_trade_api.polygon': ['py.typed'],
```
haxdds, line 40: Hi Blane-Nelson, thanks for contributing!

Why have you added a `py.typed` file for `alpaca_trade_api.polygon` package? What is the purpose?

blane-nelson: This file lets utilities (like MyPy) know that the package is using the new typing library. The files doesn't need to contain anything, it just has to be present in the package's root folder.  Refer to the  PEP 651 - https://www.python.org/dev/peps/pep-0561/ on python.org for a full explanation.

PR #410, opened 2021-03-31
ENG-4310: Add Fractional Share Support

ducille:
- Makes `qty` optional and allows fractional amounts
- Adds support for notional value orders

### DIFF
--- a/alpaca_trade_api/rest.py
+++ b/alpaca_trade_api/rest.py
@@ -269,11 +269,10 @@ class REST(object):
 
     def submit_order(self,
                      symbol: str,
-                     side: str,
-                     type: str,
-                     time_in_force: str,
                      qty: float = None,
-                     notional: float = None,
+                     side: str = "buy",
+                     type: str = "market",
+                     time_in_force: str = "day",
                      limit_price: str = None,
                      stop_price: str = None,
                      client_order_id: str = None,
@@ -282,14 +281,14 @@ class REST(object):
                      take_profit: dict = None,
                      stop_loss: dict = None,
                      trail_price: str = None,
-                     trail_percent: str = None):
+                     trail_percent: str = None,
+                     notional: float = None):
         """
         :param symbol: symbol or asset ID
+        :param qty: float. Mutually exclusive with "notional".
         :param side: buy or sell
         :param type: market, limit, stop, stop_limit or trailing_stop
         :param time_in_force: day, gtc, opg, cls, ioc, fok
-        :param qty: float. Mutually exclusive with "notional".
-        :param notional: float. Mutually exclusive with "qty".
         :param limit_price: str of float
         :param stop_price: str of float
         :param client_order_id:
@@ -302,6 +301,7 @@ class REST(object):
                {"stop_price": "297.95", "limit_price": "298.95"}
         :param trail_price: str of float
         :param trail_percent: str of float
+        :param notional: float. Mutually exclusive with "qty".
         """
         """Request a new order"""
         params = {
@@ -361,22 +361,22 @@ class REST(object):
             self,
             order_id: str,
             qty: str = None,
-            notional: str = None,
             limit_price: str = None,
             stop_price: str = None,
             trail: str = None,
             time_in_force: str = None,
-            client_order_id: str = None
+            client_order_id: str = None,
+            notional: str = None,
     ) -> Order:
         """
         :param order_id:
         :param qty: str of float
-        :param notional: str of float
         :param limit_price: str of float
         :param stop_price: str of float
         :param trail: str of float, represents trailing_price or
                trailing_percent. determined by the original order.
         :param time_in_force: day, gtc, opg, cls, ioc, fok
+        :param notional: str of float
 
         note: you cannot replace type of order. so, it was trailing_stop(e.g)
               it will remain trailing_stop.
--- a/alpaca_trade_api/rest.py
+++ b/alpaca_trade_api/rest.py
@@ -366,17 +366,15 @@ class REST(object):
             trail: str = None,
             time_in_force: str = None,
             client_order_id: str = None,
-            notional: str = None,
     ) -> Order:
         """
         :param order_id:
-        :param qty: str of float
+        :param qty: str of int
         :param limit_price: str of float
         :param stop_price: str of float
         :param trail: str of float, represents trailing_price or
                trailing_percent. determined by the original order.
         :param time_in_force: day, gtc, opg, cls, ioc, fok
-        :param notional: str of float
 
         note: you cannot replace type of order. so, it was trailing_stop(e.g)
               it will remain trailing_stop.
@@ -384,8 +382,6 @@ class REST(object):
         params = {}
         if qty is not None:
             params['qty'] = qty
-        if notional is not None:
-            params['notional'] = notional
         if limit_price is not None:
             params['limit_price'] = FLOAT(limit_price)
         if stop_price is not None:
--- a/alpaca_trade_api/rest.py
+++ b/alpaca_trade_api/rest.py
@@ -273,6 +273,7 @@ class REST(object):
                      type: str,
                      time_in_force: str,
                      qty: float = None,
+                     notional: float = None,
                      limit_price: str = None,
                      stop_price: str = None,
                      client_order_id: str = None,
@@ -287,7 +288,8 @@ class REST(object):
         :param side: buy or sell
         :param type: market, limit, stop, stop_limit or trailing_stop
         :param time_in_force: day, gtc, opg, cls, ioc, fok
-        :param qty: float
+        :param qty: float. Mutually exclusive with "notional".
+        :param notional: float. Mutually exclusive with "qty".
         :param limit_price: str of float
         :param stop_price: str of float
         :param client_order_id:
@@ -310,6 +312,8 @@ class REST(object):
         }
         if qty is not None:
             params['qty'] = qty
+        if notional is not None:
+            params['notional'] = notional
         if limit_price is not None:
             params['limit_price'] = FLOAT(limit_price)
         if stop_price is not None:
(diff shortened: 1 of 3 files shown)


alpaca_trade_api/rest.py
```diff
@@ -284,10 +285,11 @@ def submit_order(self,
                      trail_percent: str = None):
         """
         :param symbol: symbol or asset ID
-        :param qty: int
         :param side: buy or sell
         :param type: market, limit, stop, stop_limit or trailing_stop
         :param time_in_force: day, gtc, opg, cls, ioc, fok
+        :param qty: float. Mutually exclusive with "notional".
+        :param notional: float. Mutually exclusive with "qty".
```
ducille, line 292: The mutual exclusivity of `qty` and `notional` is captured in the [Orders docs](https://alpaca.markets/docs/api-documentation/api-v2/orders/), so not sure if we need to mention it here.

camelpac: I would document it. it will help to users avoiding this issue. many users just use example code as their baseline

alpaca_trade_api/rest.py
```diff
@@ -269,10 +269,11 @@ def list_orders(self,
 
     def submit_order(self,
                      symbol: str,
-                     qty: int,
```
camelpac, line 272: when doing this (changing the order of arguments), it will break any previous usage of this package. many users and packages (e.g backtrader, pylivetrader) us it like this:
`api.submit_order("AAPL", 100, "buy", "market", "day")`
not saying we can't. just something to consider (and definitely document if we do decide to do that as a breaking change)

camelpac:
also, you should probably add example usage to the readme

alpaca_trade_api/rest.py
```diff
@@ -304,11 +306,14 @@ def submit_order(self,
...
             'symbol':        symbol,
-            'qty':           qty,
             'side':          side,
             'type':          type,
             'time_in_force': time_in_force
         }
+        if qty is not None:
+            params['qty'] = qty
+        if notional is not None:
+            params['notional'] = notional
```
bala813, line 316: Add validation to make sure only one is set ? I see other parameters don't have any validation here but since this is a new breaking change, it would be good to catch early.

alpaca_trade_api/rest.py
ducille: @camelpac that's a good point. If we want to make `qty` optional while maintaining the current order of arguments, we could consider one of these approaches:

1. Make `qty` optional, leave `qty` in its current position, then add `notional` as the last kwarg. This maintains order at the expense of presenting the required `side`, `type`, and `time_in_force` arguments as optional.
2. Rename `qty` to something like `size`, then add `size_type` as the last kwarg, which accepts `"qty"` or `"notional"` and defaults to `"qty"`. This sidesteps the issue of making `side`, `type`, and `time_in_force` optional, at the expense of possibly breaking code that passes an explicitly named `qty` argument to `submit_order`.

What do you think?

camelpac: option 2 is not good if I understand you correctly.. you want to have both qty and size? that's confusing. 
option 1 is the right choice. BUT side, type and time_in_force are not "optional" they will just have a default value. optional implies they are not mandatory.. which is not correct. so it should look like this:

```py
def submit_order(self,
                 symbol: str,
                 qty: float = None,
                 side: str = "buy",
                 type: str = "market",
                 time_in_force: str = "day",
                 limit_price: str = None,
                 stop_price: str = None,
                 client_order_id: str = None,
                 extended_hours: bool = None,
                 order_class: str = None,
                 take_profit: dict = None,
                 stop_loss: dict = None,
                 trail_price: str = None,
                 trail_percent: str = None, 
                 notational: float = None)
```

ducille: Thanks for your feedback @camelpac! To clarify, my concern about making `side`, `type`, and `time_in_force` _appear_  optional--even though they're not--was rooted in the mistaken assumption that we'd need to give them null default values to accommodate changes to `qty` without making `submit_order` too "opinionated" , so to speak. Since this seems to be a non-issue (i.e., there are reasonable default values for these arguments) I agree that Option 1 makes the most sense.

FWIW, this was what I had in mind for Option 2:

```py
def submit_order(
    self,
    symbol: str,
    size: float,  # abstracted from `qty`
    side: str,
    type: str,
    time_in_force: str,
    limit_price: str = None,
    stop_price: str = None,
    client_order_id: str = None,
    extended_hours: bool = None,
    order_class: str = None,
    take_profit: dict = None,
    stop_loss: dict = None,
    trail_price: str = None,
    trail_percent: str = None, 
    size_type: str = "qty",  # can be "qty" or "notional"
)
```
```

File: alpaca_trade_api/rest.py (hunk 3 above)
Comment (camelpac) [reply alpaca_trade_api/rest.py:330]: this validation is ok, but not mandatory. if the user mis-use this api - the server will let them know. 
think about it - this mistake happens once BUT this validation happens every time the user do submit_order.
I know we are not a real time api, but this does add latency (every time).

Comment (umitanuki) [reply alpaca_trade_api/rest.py:330]: The practice of this package has been a simple wrapper of the API and to rely on the API for validation unless it's the language specific checks. I would leave it to the API side to check to avoid unnecessary inconsistency between checks in the client side and API side.

File: alpaca_trade_api/rest.py (hunk 2 above)
Comment (umitanuki) [reply alpaca_trade_api/rest.py:272]: Totally agree on the decision here for option 1. Another consideration would be to have a bail-out option to be exact the same behavior if one configures (e.g. environment variable) just in case. This is the first time breaking change on this most important method and it makes sense to pay extra coverage like that.

File: alpaca_trade_api/rest.py (hunk 3 above)
Comment (ducille) [reply alpaca_trade_api/rest.py:316]: Thanks for the context here! I'll strip out the validation and leave it to the API.

File: alpaca_trade_api/rest.py (hunk 2 above)
Comment (ducille) [reply alpaca_trade_api/rest.py:272]: > Another consideration would be to have a bail-out option to be exact the same behavior if one configures (e.g. environment variable) just in case.

@umitanuki did you have something like this in mind?

```py
try:
    resp = self.post('/orders', params)
    return self.response_wrapper(resp, Order)
except APIError as err:
    # ensure `qty` is non-null to avoid infinite recursion
    if (qty is not None) and os.environ.get("RESUBMIT_BAD_ORDERS"):
        # exclude notional parameter and resubmit order
        return self.submit_order(
            symbol,
            qty,
            side,
            type,
            time_in_force,
            limit_price,
            stop_price,
            client_order_id,
            extended_hours,
            order_class,
            take_profit,
            stop_loss,
            trail_price,
            trail_percent,
        )
    raise err
```

SC4RECOIN: option 1 looks good to me.. not sure what you mean by breaking change @umitanuki?

alpaca_trade_api/rest.py
```diff
@@ -360,23 +365,27 @@ def replace_order(
             stop_price: str = None,
             trail: str = None,
             time_in_force: str = None,
-            client_order_id: str = None
+            client_order_id: str = None,
+            notional: str = None,
     ) -> Order:
         """
         :param order_id:
-        :param qty: str of int
```
ducille, line 367: It doesn't seem that the Orders API supports replacement with fractional quantities or notional amounts (see [gobroker/oms/orderapi.go#L210-213](https://github.com/alpacahq/gobroker/blob/765a9fdbc1c81fd7abe79b7c317afff8c2c75a45/oms/orderapi.go#L210)), so I'm gonna revert this.

alpaca_trade_api/rest.py
```diff
@@ -360,23 +365,27 @@ def replace_order(
...
+        :param notional: str of float
 
         note: you cannot replace type of order. so, it was trailing_stop(e.g)
               it will remain trailing_stop.
         """
         params = {}
         if qty is not None:
             params['qty'] = qty
+        if notional is not None:
+            params['notional'] = notional
```
ducille, line 388: It doesn't seem that the Orders API supports replacement with fractional quantities or notional amounts (see [gobroker/oms/orderapi.go#L210-213](https://github.com/alpacahq/gobroker/blob/765a9fdbc1c81fd7abe79b7c317afff8c2c75a45/oms/orderapi.go#L210)), so I'm gonna revert this.

alpaca_trade_api/rest.py
raja: I don't think introducing the env var makes sense because it still introduces a breaking change. better to just bump the minor version (instead of patch) since this doesn't break current usage but adds functionality

tests/test_rest.py
```diff
@@ -820,6 +825,124 @@ def callback_429(request, context):
...
+'''
+    )
+
+    try:
+        api_v2.submit_order(
+            symbol='AAPL',
+            side='buy',
+            type='market',
+            time_in_force='day',
+
```
kalmant, line 845: minor: unnecessary empty line

alpaca_trade_api/rest.py
```diff
@@ -269,10 +269,10 @@ def list_orders(self,
 
     def submit_order(self,
                      symbol: str,
-                     qty: int,
-                     side: str,
-                     type: str,
-                     time_in_force: str,
+                     qty: float = None,
```
kalmant, line 272: If a user had a variable typed as `int` for input I think there may be some type checkers that may throw an error/warning since this parameter now has a different type than before. 
Not sure how big of an issue this is tbh (or whether this really is the case for some type checkers e.g. mypy).

ducille: Not sure about other type checkers, but mypy recognizes [duck type compatibility](https://mypy.readthedocs.io/en/latest/duck_type_compatibility.html), so it treats `int` objects as valid `float` objects

umitanuki (approved):

PR #425, opened 2021-05-03
Make changes to long-short.py to make it more Pythonic and beginner friendly

Sciguymjm:
As an example, it was destructive for a beginner to the Python language. This does not change the core design which is still flawed, but it allows the code to run fully.

Summary of changes:
- 4 spaces instead of 2 (PEP8: https://www.python.org/dev/peps/pep-0008/)
- camelCase to snake_case (PEP8)
- Remove parentheses from if statements (PEP8)
- Add some type hints where useful (syntax)
- Avoid using un-imported pandas (pd) library to create timestamps, instead use proper datetime.date (syntax)
- Remove all threading usage as it is all unnecessary when threads are immediately joined (design)
- Remove all passing of lists to functions to retrieve values and instead use return statements for said values (design)
- Add URL wrapping on API initialization to comply with the type hints of the library (misc)

This is by no means a complete fix of all issues.

### DIFF
--- a/examples/long-short.py
+++ b/examples/long-short.py
@@ -1,8 +1,9 @@
 import datetime
-import threading
+import time
+from typing import List
 
 import alpaca_trade_api as tradeapi
-import time
+from alpaca_trade_api.common import URL
 from alpaca_trade_api.rest import TimeFrame
 
 API_KEY = "YOUR_API_KEY_HERE"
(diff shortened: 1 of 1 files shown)


examples/long-short.py
```diff
@@ -11,328 +12,275 @@
...
+                # Close all positions when 15 minutes til market close.
+                print("Market closing soon.  Closing positions.")
+
+                positions = self.alpaca.list_positions()
+                for position in positions:
+                    if position.side == 'long':
+                        order_side = 'sell'
+                    else:
+                        order_side = 'buy'
+                    qty = abs(int(float(position.qty)))
```
gnvk, line 68: Can you please explain why we need the float conversion here?

examples/long-short.py
```diff
@@ -11,328 +12,275 @@
...
+            return True
+
+    # Get percent changes of the stock prices over the past 10 minutes.
+    def get_percent_changes(self) -> None:
+        length = 10
+        for i, stock in enumerate(self.all_stocks):
+            bars = self.alpaca.get_bars(stock[0], TimeFrame.Minute,
+                                        datetime.date.today().isoformat(),
+                                        datetime.date.today().isoformat(), limit=length,
+                                        adjustment='raw')
```
gnvk, line 272: This will return the first 10 minute bars of today for the given symbol, instead of the last 10 minutes. I know it's also broken in the original but please fix it.

examples/long-short.py
```diff
@@ -11,328 +12,275 @@
...
+            self.all_stocks[i][1] = (bars[stock[0]][len(bars[stock[0]]) - 1].c -
```
gnvk, line 273: [minor] You can simply use -1
```suggestion
            self.all_stocks[i][1] = (bars[stock[0]][-1].c -
```

examples/long-short.py
```diff
@@ -11,328 +12,275 @@
...
-          # Position not in short list either.  Clear position.
-          if(position.side == "long"):
-            side = "sell"
-          else:
-            side = "buy"
-          respSO = []
-          tSO = threading.Thread(target=self.submitOrder, args=[abs(int(float(position.qty))), position.symbol, side, respSO])
-          tSO.start()
-          tSO.join()
+            if self.long.count(position.symbol) == 0:
```
gnvk, line 107: [minor]
```suggestion
            if position.symbol not in self.long.count(position.symbol):
```

examples/long-short.py
gnvk: There are a lot more similar double `int(float(...))` conversions throughout the file.

examples/long-short.py
```diff
@@ -11,328 +12,275 @@
...
+        equity = int(float(self.alpaca.get_account().equity))
+
+        self.short_amount = equity * 0.30
+        self.long_amount = equity - self.short_amount
+
+        resp_get_tp_long = self.get_total_price(self.long)
+
+        resp_get_tp_short = self.get_total_price(self.short)
+
+        self.q_long = int(self.long_amount // resp_get_tp_long)
```
gnvk, line 222: [minor] no need for the int conversion after //

examples/long-short.py
```diff
@@ -11,328 +12,275 @@
...
+    def get_total_price(self, stocks) -> float:
+        total_price = 0
+        for stock in stocks:
+            bars = self.alpaca.get_bars(stock, TimeFrame.Minute,
+                                        datetime.date.today().isoformat(),
+                                        datetime.date.today().isoformat(), limit=1,
+                                        adjustment='raw')
+
+            total_price += bars[stock][0].c
+        return total_price
```
gnvk, line 235: While this logic is correct, I think using the latest trade price is much more accurate. The price can move a lot since the last minute bars closing price.
```suggestion
        return sum(self.alpaca.get_latest_trade(stock).price for stock in stocks)
```

examples/long-short.py
```diff
@@ -11,328 +12,275 @@
...
+                            side = "sell"
+                        else:
+                            # Too little long positions.  Buy some more.
+                            side = "buy"
+                        self.submit_order(abs(diff), position.symbol, side)
+                    executed[0].append(position.symbol)
+                    self.blacklist.add(position.symbol)
+
+        # Send orders to all remaining stocks in the long and short list.
+        resp_send_bo_long = self.send_batch_order(self.q_long, self.long, "buy")
```
gnvk, line 161: [minor] tuple unpacking would be much nicer:
```suggestion
        executed, incomplete = self.send_batch_order(self.q_long, self.long, "buy")
```

gnvk (commented):
This PR is **amazing**, thank you very much! 🥇 

I agree it's unfortunately not complete, but it's a huge step to the right direction. Still, I left some comments on issues I feel should be included, especially the ones that make the implementation incorrect (e.g. querying the first 10 minute bars of the day instead of the latest 10 minute bars).

PR #435, opened 2021-05-25
Add daily bars to data stream

gnvk:


alpaca_trade_api/stream.py
```diff
@@ -136,7 +148,7 @@ def _subscribe(self, handler, symbols, handlers):
         if self._running:
             asyncio.get_event_loop().run_until_complete(self._subscribe_all())
 
-    async def _unsubscribe(self, trades=(), quotes=(), bars=()):
+    async def _unsubscribe(self, trades=(), quotes=(), bars=(), daily_bars=()):
```
kalmant, line 151: Please add `daily_bars` to the message that is sent

gnvk: Great catch, thank you!

kalmant (approved):
LGTM

PR #439, opened 2021-06-02
Update the websockets package

camelpac:
The websockets package as a new major version. 
After reading the release notes and making sure nothing breaks, I suggest increasing allowed version of the package.

websocket-client 1.0 dropped support for python2. but we're ok with that.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -3,5 +3,5 @@ numpy
 requests>2,<3
 urllib3>1.24,<2
 websocket-client>=0.56.0,<1
-websockets>=8.0,<9
+websockets>=8.0,<10
 msgpack==1.0.2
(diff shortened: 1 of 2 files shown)


requirements/requirements.txt
```diff
@@ -2,6 +2,6 @@ pandas
 numpy
 requests>2,<3
 urllib3>1.24,<2
-websocket-client>=0.56.0,<1
-websockets>=8.0,<9
+websocket-client>=0.56.0,<2
+websockets>=8.0,<10
```
gnvk, line 6: In https://github.com/alpacahq/alpaca-trade-api-python/pull/447 it says:
pyup.io/safety-ci — websockets 8.1 has known security vulnerabilities.

Shouldn't we go with a higher minimum version because of this?

camelpac: if we do, we will prevent users that may use other packages that require that version (8.1 was the default version for a very long time).
We could recommend using a higher version, but I don't think we should prevent it

camelpac: added this note and merging: 
image

PR #443, opened 2021-06-08
Fix example code to work with data v2

camelpac:
make the example code compatible with data v2

### DIFF
--- a/examples/overnight_hold.py
+++ b/examples/overnight_hold.py
@@ -31,22 +31,30 @@ def get_ratings(symbols, algo_time):
     formatted_time = None
     if algo_time is not None:
         # Convert the time to something compatable with the Alpaca API.
+        start_time = (algo_time.date() -
+                      timedelta(days=window_size)).strftime(api_time_format)
         formatted_time = algo_time.date().strftime(api_time_format)
     while index < len(assets):
         symbol_batch = [
             asset.symbol for asset in assets[index:index+batch_size]
         ]
         # Retrieve data for this batch of symbols.
-        barset = {}
-        for symbol in symbol_batch:
-            bars = api.get_bars(symbol,
-                                TimeFrame.Day,
-                                formatted_time - timedelta(days=window_size),
-                                formatted_time,
-                                limit=window_size,
-                                adjustment='raw')
-            barset[symbol] = bars
-
+        # barset = {}
+        # for symbol in symbol_batch:
+        #     bars = api.get_bars(symbol,
+        #                         TimeFrame.Day,
+        #                         start_time,
+        #                         formatted_time,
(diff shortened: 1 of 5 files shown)


examples/overnight_hold.py
```diff
@@ -31,22 +31,30 @@ def get_ratings(symbols, algo_time):
...
-        for symbol in symbol_batch:
-            bars = api.get_bars(symbol,
-                                TimeFrame.Day,
-                                formatted_time - timedelta(days=window_size),
-                                formatted_time,
-                                limit=window_size,
-                                adjustment='raw')
-            barset[symbol] = bars
-
+        # barset = {}
```
gnvk, line 42: I think these should be removed, not just commented out.

examples/overnight_hold.py
```diff
@@ -160,16 +168,21 @@ def get_value_of_assets(api, shares_bought, on_date):
...
+    )
+    # barset = {}
+    # for symbol in shares_bought.keys():
+    #     bars = api.get_bars(symbol,
+    #                         TimeFrame.Day,
+    #                         on_date.date(),
+    #                         on_date.date(),
+    #                         limit=1,
+    #                         adjustment='raw')
+    #     barset[symbol] = bars
```
gnvk, line 185: Same here

examples/overnight_hold.py
```diff
@@ -31,22 +31,30 @@ def get_ratings(symbols, algo_time):
...
-            bars = api.get_bars(symbol,
-                                TimeFrame.Day,
-                                formatted_time - timedelta(days=window_size),
-                                formatted_time,
-                                limit=window_size,
-                                adjustment='raw')
-            barset[symbol] = bars
-
+        # barset = {}
+        # for symbol in symbol_batch:
```
gnvk, line 43: It took quite some time even for me to figure out that the section that was commented out was v2 and the original v1.

camelpac: thing is that get_barset is no longer documented but still works. 
it's more efficient - it gets 200 symbols at once.
which means it's faster.
I added v2 so we could easily switch when the undocumented api stops working.

if you'd like, I could add a comment describing this.

gnvk: - the v1 (get_barset) will be deprecated
- please add a comment
- you could use multithreading to fetch the symbols concurrently instead of sequentially

gnvk: the absolute best solution however would be to rewrite this (and the other) examples to use the streaming api when necessary!

camelpac: I will add this note and merge it.
using the websocket is good for live trading but this example also provides a backtester so we must have historical data.
I will add threading in the future, but that also has limits

PR #446, opened 2021-06-10
Add support for closing partial position

camelpac:
Add support for this API: https://alpaca.markets/docs/api-documentation/api-v2/positions/#close-a-position.
According to the docs one could specify partial fractional qty or percentage. 

percentage seems to be currently unsupported, so it is commented out and will be commented in once api supports it.

For now, one could close partial fractional qty like this: 

```py
api.close_position("AAPL", 0.5)
```

Read the docs for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,10 @@ The solution - manually install these package before installing alpaca-trade-api
 ```bash
 pip install pandas==1.1.5 numpy==1.19.4 scipy==1.5.4
 ```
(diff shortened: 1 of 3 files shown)


gnvk (approved):
LGTM

camelpac:
thanks. I will wait for the PR with websockets version to be merged, then merge this once since the pyup check doesn't pass with websokcets 8.1

PR #448, opened 2021-06-13
Fix account typo

JohnStrunk226:
Fixed "quotes and accoutn updates" to be "quotes and account updates" in the readme. A simple typo I found while reading through it.


camelpac:
thanks

PR #451, opened 2021-06-17
Try to consume again when we get a connection error 

camelpac:
3 retries by default

the way it was - there was never a retry, just a loop that runs forever if we get 1 connection error


gnvk (approved):
LGTM

PR #452, opened 2021-06-18
Extend list_orders functionality with symbols filtering

camelpac:
https://alpaca.markets/docs/api-documentation/api-v2/orders/#get-a-list-of-orders indicates you could use a list of symbols as filters. adding that as a param.

You could now run this code:

```py
api.list_orders(status="closed", symbols=["GOOG", "TSLA", "MSFT"])
```


gnvk (approved):
LGTM

PR #456, opened 2021-06-22
Allow all versions of pandas and numpy

camelpac:
instruct pyup to ignore the version and stop opening PRs for updating these packages.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,5 +1,5 @@
-pandas
-numpy
+pandas  # pyup: ignore - allow all versions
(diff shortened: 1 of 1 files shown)


gnvk (commented):
I agree with the pyup annotation, but are you convinced we don't need a minimum? I'm pretty sure the SDK won't work with very old pandas / numpy versions.

camelpac:
pylivetrader is using python 3.6. this is why we have this requirement

gnvk (approved):
LGTM

PR #461, opened 2021-07-02
Fix martingale example to use correct var names

thealmightygrant:
The stream is named conn, not stream in this example, and the example does not work currently because of it.


gnvk (approved):
Thank you for the fix!

PR #464, opened 2021-07-06
Add LULD support to data stream

kalmant:


alpaca_trade_api/stream.py
```diff
@@ -276,8 +303,11 @@ async def _run_forever(self):
                         int(os.environ.get('APCA_RETRY_WAIT', 3)))
                 log.warn('websocket error, restarting connection: ' +
                          str(wse))
+            except Exception as e:
```
gnvk, line 306: This is not related to the LULD messages. Also, I usually dislike handling all exceptions like this, though only logging them could be acceptable.

kalmant: The only reason I added this was that during my LULD testing I ran into some exceptions that were not logged/reported at all so even finding out about their existence was difficult.

gnvk (approved):
Ship it

Issue #407, opened 2021-03-24
run_forever()'s restart condition never met


Chrisrdouglas:
in this loop on line 208 of stream.py, self._running is never set to false in the exception. retries starts at 0, then incremented in the exception. This has it compare 1>3 which is false. This means that self._running is never set to false. Then in finally block on line 226, the break statement on line 228 will never be reached. Wouldn't this stop it from restarting on the next pass of the loop?
```python
        while True: # stream.py - line 208
            try:
                if not self._running:
                    await self._start_ws()
                    self._running = True
                    retries = 0
                    await self._consume()
            except websockets.WebSocketException as wse:
                retries += 1
                if retries > int(os.environ.get('APCA_RETRY_MAX', 3)):
                    await self.close()
                    self._running = False
                    raise ConnectionError("max retries exceeded")
                if retries > 1:
                    await asyncio.sleep(
                        int(os.environ.get('APCA_RETRY_WAIT', 3)))
                log.warn('websocket error, restarting connection: ' +
                         str(wse))
            finally:
                print(retries)
                if not self._running:
                    break
                await asyncio.sleep(0.01)```

Chrisrdouglas:
in this loop on line 208 of stream.py self._running is never set to false in the exception
`        while True:
            try:
                if not self._running:
                    await self._start_ws()
                    self._running = True
                    retries = 0
                    await self._consume()
            except websockets.WebSocketException as wse:
                retries += 1
                if retries > int(os.environ.get('APCA_RETRY_MAX', 3)):
                    await self.close()
                    self._running = False
                    raise ConnectionError("max retries exceeded")
                if retries > 1:
                    await asyncio.sleep(
                        int(os.environ.get('APCA_RETRY_WAIT', 3)))
                log.warn('websocket error, restarting connection: ' +
                         str(wse))
            finally:
                print(retries)
                if not self._running:
                    break
                await asyncio.sleep(0.01)`
Wouldn't this stop it from restarting on the next pass of the loop?

nickvertucci:
@Chrisrdouglas Would this help solve 1006 websocket errors and reconnect/resubscribe?

Chrisrdouglas:
@nickvertucci

hmm, I haven't looked at this code since march but I doubt it would solve the issue of getting disconnected because of that internal connection issue. Are a lot of people having that problem?

noahkuhn:
[...]

PR #466, opened 2021-07-09
restart websocket on ConnectionClosedError

srgustafson8:
As mentioned in #407, the reconnect logic in `_run_forever()` never actually reconnected to the Websocket in case of a connection error as `_running` was only set to False when max retries were hit.

This PR adds a check for if the exception is a ConnectionClosedError, and then attempts a full reconnection. It also brings the trading stream to match the data stream.

### DIFF
--- a/alpaca_trade_api/stream.py
+++ b/alpaca_trade_api/stream.py
@@ -258,9 +258,13 @@ class DataStream:
         log.info('started data stream')
         retries = 0
         self._running = False
+        self._conn_error = False
         while True:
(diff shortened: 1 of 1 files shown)


gnvk:
I believe this issue was fixed by https://github.com/alpacahq/alpaca-trade-api-python/commit/d974797e12856414e978db5346d027df5433a608. @srgustafson8 can you please confirm it? If so, please close the PR. If not, update it. Thank you!

PR #474, opened 2021-07-30
Add SMA and ADTV

kalmant:
Once `feature/luld` gets merged/updated this PR's base should also be updated.


umitanuki:
We understand it would be nice to have indicator calculations in this repository but look forward to a complete PR and close this for now

PR #483, opened 2021-08-30
Add support for async rest

camelpac:
In this PR we add ability to use asyncio python to use the historic endpoints.


alpaca_trade_api/rest_async.py
```diff
@@ -0,0 +1,183 @@
+import os
+import aiohttp
```
gnvk, line 2: aiohttp is missing from the requirements.txt!

alpaca_trade_api/rest_async.py
```diff
@@ -0,0 +1,183 @@
...
+NY = 'America/New_York'
+
+
+class AsyncRest:
+    def __init__(self,
+                 key_id: str = None,
+                 secret_key: str = None,
+                 data_url: URL = None,
+                 api_version: str = None,
+                 oauth=None,
```
gnvk, line 18: [minor] unused?

alpaca_trade_api/rest_async.py
```diff
@@ -0,0 +1,183 @@
...
+                 raw_data: bool = False
+                 ):
+        """
+        :param raw_data: should we return api response raw or wrap it with
+                         Entity objects.
+        """
+        self._key_id, self._secret_key = key_id, secret_key
+        self._data_url: URL = URL(data_url or get_data_url())
+        self._api_version = get_api_version(api_version)
```
gnvk, line 27: [minor] unused?

alpaca_trade_api/rest_async.py
```diff
@@ -0,0 +1,183 @@
...
+        self._use_raw_data = raw_data
+        self._retry = int(os.environ.get('APCA_RETRY_MAX', 3))
+        self._retry_wait = int(os.environ.get('APCA_RETRY_WAIT', 3))
```
gnvk, line 30: All of these seems to be unused.

alpaca_trade_api/rest_async.py
```diff
@@ -0,0 +1,183 @@
...
+            # human error to put non-SSL endpoint could exploit
+            # uncanny issues in non-GET request redirecting http->https.
+            # It's better to fail early if the URL isn't right.
+            'allow_redirects': False,
+        }
+        opts['params'] = payload
+
+        return opts
+
+    async def _request(self, symbol, url, payload):
```
gnvk, line 152: [minor] symbol is unused

alpaca_trade_api/rest_async.py
```diff
@@ -0,0 +1,183 @@
...
+                    async with session.get(url, **opts) as response:
+
+                        response = await response.json()
+                        page_token = response.get('next_page_token')
+                        payload["page_token"] = page_token
+                        yield response
+
+                        if not page_token:
+                            break
+            except Exception as e:
```
gnvk, line 166: I really don't like handling all exceptions like this.

examples/historic_async.py
```diff
@@ -0,0 +1,126 @@
...
+import pandas as pd
+
+NY = 'America/New_York'
+
+
+async def get_historic_bars(symbols, start, end, timeframe: TimeFrame):
+    major = sys.version_info.major
+    minor = sys.version_info.minor
+    if major < 3 or minor < 6:
+        raise Exception('asyncio is not support in your python version')
```
gnvk, line 17: In my opinion this check should go to the SDK.

examples/historic_async.py
```diff
@@ -0,0 +1,126 @@
...
+    end = pd.Timestamp('2021-08-30', tz=NY).date().isoformat()
+    timeframe: TimeFrame = TimeFrame.Day
+    await get_historic_bars(symbols, start, end, timeframe)
+    await get_historic_trades(symbols, start, end, timeframe)
+    await get_historic_quotes(symbols, start, end, timeframe)
+
+
+if __name__ == '__main__':
+    import time
+    import yaml
```
gnvk, line 104: Not in requirements.txt

examples/historic_async.py
```diff
@@ -0,0 +1,126 @@
...
+
+    with open("./config.yaml", mode='r') as f:
```
gnvk, line 106: An example should be given for this.

alpaca_trade_api/rest_async.py
```diff
@@ -0,0 +1,183 @@
...
+                if len(df) >= limit:
+                    break
+
+        return df
+
+    async def get_bars_async(self, symbol, start, end, timeframe, limit=1000):
+        _type = "bars"
+
+        payload = {
+            "adjustment": 'raw',
```
gnvk, line 67: [minor] This should be an argument instead

examples/historic_async.py
```diff
@@ -0,0 +1,126 @@
...
+    for symbol in symbols:
+        tasks.append(
+            rest.get_bars_async(symbol, start, end, timeframe, limit=500))
+    if minor >= 8:
+        results = await asyncio.gather(*tasks)
+    else:
+        results = await gather_with_concurrency(500, *tasks)
+
+    bad_requests = 0
+    for response in results:
```
gnvk, line 32: It's not trivial how to actually consume the responses (other than printing their count). E.g. how to pair the symbol with the dataframe.

examples/historic_async.py
```diff
@@ -0,0 +1,126 @@
...
+
+    tasks = []
+
+    for symbol in symbols:
+        tasks.append(
+            rest.get_bars_async(symbol, start, end, timeframe, limit=500))
+    if minor >= 8:
+        results = await asyncio.gather(*tasks)
+    else:
+        results = await gather_with_concurrency(500, *tasks)
```
gnvk, line 29: This too could be handled on the SDK side.

examples/historic_async.py
```diff
@@ -0,0 +1,126 @@
...
+                     secret_key=api_secret)
+    NY = 'America/New_York'
+
+    api = tradeapi.REST(key_id=api_key_id,
+                        secret_key=api_secret,
+                        base_url=URL(base_url))
+
+    start_time = time.time()
+    loop = asyncio.get_event_loop()
+    symbols = [el.symbol for el in api.list_assets()]
```
gnvk, line 123: It would make sense to only list active assets
```suggestion
    symbols = [el.symbol for el in api.list_assets(status='active')]
```

examples/historic_async.py
```diff
@@ -0,0 +1,126 @@
+import sys
+import asyncio
+import alpaca_trade_api as tradeapi
+from alpaca_trade_api.rest import TimeFrame, URL
+from alpaca_trade_api import rest_async
```
gnvk, line 5: unused

examples/historic_async.py
```diff
@@ -0,0 +1,126 @@
...
+from alpaca_trade_api.rest_async import gather_with_concurrency
+from alpaca_trade_api import AsyncRest
```
gnvk, line 7: this import is incorrect, `AsyncRest` is in `alpaca_trade_api.rest_async`

examples/historic_async.py
```diff
@@ -0,0 +1,126 @@
...
+import pandas as pd
+
+NY = 'America/New_York'
```
gnvk, line 10: unused

alpaca_trade_api/rest_async.py
```diff
@@ -0,0 +1,183 @@
+import os
+import aiohttp
+import asyncio
+from alpaca_trade_api.entity_v2 import BarsV2, QuotesV2, TradesV2, \
+    EntityList, TradeV2, QuoteV2
+import pandas as pd
+from alpaca_trade_api.common import URL, get_api_version, get_data_url
+
+NY = 'America/New_York'
```
gnvk, line 9: unused

alpaca_trade_api/rest_async.py
camelpac: added

alpaca_trade_api/rest_async.py
camelpac: removed

alpaca_trade_api/rest_async.py
camelpac: removed

alpaca_trade_api/rest_async.py
camelpac: removed

alpaca_trade_api/rest_async.py
camelpac: the thing is that this method is used for bulk data retrieve. if I want to get data for 2000 symbols, if one fails I don't want to fail everything. I want to provide best effort.
for instance, if I remove this in my example when waiting for tasks to complete (in this code: `results = await asyncio.gather(*tasks)`

I get this exception for one symbol: 

image

that fails everything.

do you have a better suggestion to allow getting the "best effort" outcome?

examples/historic_async.py
camelpac: in order to do that I need to move `get_historic_bars(symbols, start, end, timeframe: TimeFrame)` inside the sdk. 
right now I place it in the example code and this is used to get data for multiple symbols. 

do you think getting data for multiple symbols should sit inside the sdk?

(btw, this issue was very hard to understand on why I'm getting different results on different python versions)

examples/historic_async.py
camelpac: added

examples/historic_async.py
camelpac: added

alpaca_trade_api/rest_async.py
camelpac: added

examples/historic_async.py
camelpac: now returning the symbol with the dataframe

examples/historic_async.py
camelpac: this is only possible if we add the "get data for multiple symbols" to the sdk. what do you think?

examples/historic_async.py
camelpac: I actually added this to the `__init__` file:
image

so it's possible. but I changed the import to make it more readable.

examples/historic_async.py
camelpac: it's used in `async def main(symbols):` but it was defined twice. removed one.

alpaca_trade_api/rest_async.py
gnvk: I would let the caller decide how he wants to handle the exception. In this particular case I would
- add `return_exceptions=True` to the `asyncio.gather` call(s)
- when iterating through the results, check if it's an exception

examples/historic_async.py
gnvk: I understand. For now it's okey this way. In this case the only thing I'd do is I refactored the common parts of `get_historic_bars`, `get_historic_trades` and `get_historic_quotes` to a common base function.

examples/historic_async.py
gnvk: Where? I don't see it.

gnvk: Also, I think defaulting to the env vars would be nice in this example as well.

examples/historic_async.py
gnvk: I'm still missing to see it in an example.

examples/historic_async.py
gnvk: Same answer: let's stick to this for now with a refactor.

examples/historic_async.py
camelpac: in the docstring of the function
image

camelpac: now also checking the env variables
image

examples/historic_async.py
camelpac: now each function returns a tuple including the symbol

image

examples/historic_async.py
camelpac: done

camelpac (commented):
fixed all comments

alpaca_trade_api/rest_async.py
```diff
@@ -0,0 +1,180 @@
...
+            "end":        end,
+            "timeframe":  timeframe.value,
+            "limit":      limit,
+        }
+        df = await self._iterate_requests(symbol, payload, limit, _type,
+                                          BarsV2)
+
+        return symbol, df
+
+    async def get_trades_async(self, symbol, start, end, timeframe,
```
gnvk, line 77: `get_trades_async` should not have a `timeframe` argument!

alpaca_trade_api/rest_async.py
```diff
@@ -0,0 +1,180 @@
...
+            "start": start,
+            "end":   end,
+            "limit": limit,
+        }
+        df = await self._iterate_requests(symbol, payload, limit, _type,
+                                          TradesV2)
+
+        return symbol, df
+
+    async def get_quotes_async(self, symbol, start, end, timeframe,
```
gnvk, line 91: `get_quotes_async` should not have a `timeframe` argument!

examples/historic_async.py
```diff
@@ -0,0 +1,126 @@
...
+
+if __name__ == '__main__':
+    """
+    Credentials for this example is kept in a yaml config file.
+    an example to such a file:
+    
+    key_id: "<YOUR-API-KEY>"
+    secret: "<YOUR-API-SECRET>"
+    feed: iex
+    base_url: https://paper-api.alpaca.markets
```
gnvk, line 101: I would actually put this into a config.yaml file, so that the user can simply run the example if he has the proper environment variables set.

examples/historic_async.py
```diff
@@ -0,0 +1,126 @@
...
+    elif data_type == DataType.Trades:
+        return rest.get_trades_async
+    elif data_type == DataType.Quotes:
+        return rest.get_quotes_async
+    else:
+        raise Exception(f"Unsupoported data type: {data_type}")
+
+
+async def get_historic_data_base(symbols, data_type: DataType, start, end,
+                                 timeframe: TimeFrame):
```
gnvk, line 31: timeframe make no sense for trades and quotes. For the sake of this example a constant timeframe (or even the default!) for bars would be fine.

gnvk (approved):
LGTM

PR #488, opened 2021-09-13
Add crypto marketdata

gnvk:


kalmant (approved):
LGTM

camelpac (commented):
Looks good. The crypto data stream class and the stocks data stream class have a lot of shared code. I would consider putting the shared code in the base class (all the subscribe methods)

camelpac (approved):
LGTM

PR #493, opened 2021-09-15
Support multiple symbols in data queries

gnvk:


alpaca_trade_api/rest.py
```diff
@@ -546,7 +546,7 @@ def get_last_quote(self, symbol: str) -> Quote:
 
     def _data_get(self,
                   endpoint: str,
-                  symbol: str,
+                  symbol: Union[str, List[str]],
```
camelpac, line 549: symbol is probably not the best name for a list.
symbols is not the best too but better than symbol.

alpaca_trade_api/rest.py
```diff
@@ -562,21 +562,31 @@ def _data_get(self,
...
+                for item in resp.get(endpoint, []) or []:
+                    yield item
+                    total_items += 1
+            else:
+                by_symbol = resp.get(endpoint, {}) or {}
+                for sym, items in sorted(by_symbol.items()):
+                    for item in items or []:
+                        item['S'] = sym
+                        yield item
+                        total_items += 1
```
camelpac, line 581: so now if we limit to 1000 and we ask for 1000 symbols we get 1 item for each? shouldn't it be 1000 for each symbol?

alpaca_trade_api/rest.py
```diff
@@ -588,22 +598,19 @@ def get_trades_iter(self,
...
-                                           raw=True))
+                                           start, end, limit, raw=True))
         return TradesV2(trades)
 
     def get_quotes_iter(self,
-                        symbol: str,
-                        start: str,
-                        end: str,
+                        symbol: Union[str, List[str]],
+                        start: Optional[str] = None,
```
camelpac, line 612: just a note, Optional is a Union of your type and None, so if you use Optional the default value is None

camelpac (commented):
looks good, small comments.

alpaca_trade_api/rest.py
gnvk: renamed to `symbol_or_symbols`

alpaca_trade_api/rest.py
gnvk: No. The server-side multi symbol endpoints return the requested data in symbol-order first, then by timestamp. So for if we ask for 1000 symbols and set the limit to 1000, we may only get from the first symbol (if it has 1000 items in the given interval). Btw unfortunately there is a dissonance between the REST API's `limit` parameter and the SDK's `limit` parameter: the first one is actually a page size, while the second one is the actual total limit of the (automatically paginated) response size.

alpaca_trade_api/rest.py
gnvk: This is exactly what I intended.

camelpac (approved):

PR #494, opened 2021-09-15
Feature: Custom TimeRanges

### LINKED ISSUE #492: Feature Request: Customizable TimeFrame
We'd like to be able to use variable versions of the TimeFrame, to accommodate different scenarios that we require for barsets.

An ideal usage of the API would be something like this:

```python
rest_api = REST()
time_frame = TimeFrame(15, TimeFrameUnit.minutes)
result = api.get_bars("APPL", time_frame, start='2021-09-09', end='2021-09-10', limit=1000, adjustment='raw').df
```

Relates to #487 and #417

AlphaGit:
Fixes #492.

Introduces the ability to create custom timeranges for API Client consumers. Validates that the timeranges created are valid all of the time, so at no point an invalid request can be sent.


alpaca_trade_api/rest.py
```diff
@@ -69,11 +70,62 @@ def response(self):
             return self._http_error.response
 
 
-class TimeFrame(str):
-    Day = "1Day"
-    Hour = "1Hour"
-    Minute = "1Min"
-    Sec = "1Sec"
+class TimeFrameUnit(Enum):
+    Second = "Sec"
```
gnvk, line 74: The alpaca API actually doesn't support second bars.

gnvk (approved):
Nice change! I had small comment.

alpaca_trade_api/rest.py
AlphaGit: Thanks, I'll remove it. I based myself on the old `TimeFrame` enum and that sent me in the wrong direction. I'll send a fix soon.

AlphaGit: Changes applied in 9cb1b4d.

gnvk (approved):
Thank you!

PR #517, opened 2021-10-28
Security Points

0xanon:
Adding additional security like Ddos protection to alpaca and also securing it from open SSRF attacks


haxdds:
Hi 0xanon, please submit this suggestion as an issue on [the API repo](https://github.com/alpacahq/Alpaca-API).

PR #520, opened 2021-10-29
Fix code - remove unused function call

camelpac:
also add handling for keyboard interrupt.


gnvk (commented):
Is this example still relevant? I thought the stream client already restarts itself automatically.

PR #536, opened 2021-12-01
Handle trade error cancels and corrections

ccnlui:
## Context
Handle new messages as implicit subscriptions messages that come with subscribing to trades
- trade error cancels 
- trade corrections

The biggest change is letting user pass in `handler_cancel_errors` and `handler_corrections` into `subscribe_trades()`.
There is also the other option of using decorators, but it would have to be two separate calls to subscribe to trades and register error cancels / corrections handler separately.

## Test results:
```
cancel_error {'T': 'x', 'S': 'NCLH', 'i': 71692774223953, 'x': 'D', 'p': 18.365, 's': 250000, 'a': 'C', 'z': 'A', 't': Timestamp(seconds=1638387516, nanoseconds=886094000)}
corrections {'T': 'c', 'S': 'DIA', 'x': 'D', 'oi': 79372157998682, 'op': 349.15, 'os': 25000, 'oc': [' ', '7', 'V'], 'ci': 79372444317008, 'cp': 349.15, 'cs': 25000, 'cc': [' ', '7', 'Z', 'V'], 'z': 'B', 't': Timestamp(seconds=1638387766, nanoseconds=854652)}
```

### DIFF
--- a/alpaca_trade_api/entity_v2.py
+++ b/alpaca_trade_api/entity_v2.py
@@ -61,6 +61,32 @@ luld_mapping_v2 = {
     "z": "tape"
 }
 
+cancel_error_mapping_v2 = {
+    "S": "symbol",
+    "i": "id",
+    "x": "exchange",
+    "p": "price",
+    "s": "size",
+    "a": "cancel_error_action",
+    "z": "tape",
+    "t": "timestamp",
+}
+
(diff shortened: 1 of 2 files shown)


gnvk (approved):
LGTM

PR #555, opened 2022-01-10
Add news api

ccnlui:
## Context
Added support for news api, both stream and historical endpoint

## TODO
- ~Add tests~ (done)

### DIFF
--- a/alpaca_trade_api/entity_v2.py
+++ b/alpaca_trade_api/entity_v2.py
@@ -87,19 +87,6 @@ correction_mapping_v2 = {
     "t": "timestamp",
 }
 
-news_mapping_v2 = {
-    "id": "id",
-    "headline": "headline",
-    "summary": "summary",
-    "author": "author",
-    "created_at": "created_at",
-    "updated_at": "updated_at",
-    "url": "url",
-    "content": "content",
-    "symbols": "symbols",
-    "source": "source",
-}
-
 
 class EntityListType(Enum):
     Trade = Trade, trade_mapping_v2
@@ -238,9 +225,9 @@ class LatestQuotesV2(dict):
             self[k] = _convert_or_none(QuoteV2, v)
 
 
-class NewsV2(Remapped, Entity):
+class NewsV2(Entity):
     def __init__(self, raw):
-        super().__init__(news_mapping_v2, raw)
+        super().__init__(raw)
 
 
 class NewsListV2(list):
--- a/alpaca_trade_api/entity_v2.py
+++ b/alpaca_trade_api/entity_v2.py
@@ -89,16 +89,17 @@ correction_mapping_v2 = {
 
 news_mapping_v2 = {
     "id": "id",
-	"headline": "headline",
-	"summary": "summary",
-	"author": "author",
-	"created_at": "created_at",
-	"updated_at": "updated_at",
-	"url": "url",
-	"content": "content",
-	"symbols": "symbols",
-	"source": "source",
+    "headline": "headline",
+    "summary": "summary",
+    "author": "author",
+    "created_at": "created_at",
+    "updated_at": "updated_at",
+    "url": "url",
+    "content": "content",
+    "symbols": "symbols",
+    "source": "source",
 }
+
 
 class EntityListType(Enum):
     Trade = Trade, trade_mapping_v2
--- a/alpaca_trade_api/entity_v2.py
+++ b/alpaca_trade_api/entity_v2.py
@@ -87,6 +87,18 @@ correction_mapping_v2 = {
     "t": "timestamp",
 }
 
+news_mapping_v2 = {
+    "id": "id",
+	"headline": "headline",
+	"summary": "summary",
+	"author": "author",
+	"created_at": "created_at",
+	"updated_at": "updated_at",
+	"url": "url",
+	"content": "content",
+	"symbols": "symbols",
+	"source": "source",
+}
 
 class EntityListType(Enum):
     Trade = Trade, trade_mapping_v2
@@ -190,6 +202,13 @@ class CorrectionV2(Remapped, _NanoTimestamped, Entity):
 
     def __init__(self, raw):
         super().__init__(correction_mapping_v2, raw)
+
+
+class NewsV2(Remapped, _NanoTimestamped, Entity):
+    _tskeys = ('t',)
+
+    def __init__(self, raw):
+        super().__init__(news_mapping_v2, raw)
 
 
 class SnapshotV2:
--- a/alpaca_trade_api/entity_v2.py
+++ b/alpaca_trade_api/entity_v2.py
@@ -204,13 +204,6 @@ class CorrectionV2(Remapped, _NanoTimestamped, Entity):
         super().__init__(correction_mapping_v2, raw)
 
 
-class NewsV2(Remapped, _NanoTimestamped, Entity):
-    _tskeys = ('t',)
-
-    def __init__(self, raw):
-        super().__init__(news_mapping_v2, raw)
-
-
 class SnapshotV2:
     def __init__(self, raw):
         self.latest_trade = _convert_or_none(TradeV2, raw.get('latestTrade'))
@@ -242,6 +235,16 @@ class LatestQuotesV2(dict):
     def __init__(self, raw):
         for k, v in raw.items():
             self[k] = _convert_or_none(QuoteV2, v)
+
+
+class NewsV2(Remapped, Entity):
+    def __init__(self, raw):
+        super().__init__(news_mapping_v2, raw)
+
+
+class NewsListV2(list):
+    def __init__(self, raw):
+        super().__init__([NewsV2(o) for o in raw])
 
 
 def _convert_or_none(entityType, value):
--- a/alpaca_trade_api/rest.py
+++ b/alpaca_trade_api/rest.py
@@ -5,7 +5,6 @@ import requests
 from requests.exceptions import HTTPError
 import time
 from enum import Enum
-import re
 from alpaca_trade_api import __version__
 from .common import (
     get_base_url,
@@ -37,6 +36,7 @@ BarIterator = Iterator[Union[Bar, dict]]
 NewsIterator = Iterator[Union[NewsV2, dict]]
 
 DATA_V2_MAX_LIMIT = 10000  # max items per api call
+NEWS_MAX_LIMIT = 50  # max items per api call
 
 
 class RetryException(Exception):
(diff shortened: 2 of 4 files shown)


alpaca_trade_api/entity_v2.py
```diff
@@ -87,6 +87,19 @@
     "t": "timestamp",
 }
 
+news_mapping_v2 = {
```
gnvk, line 90: Isn't this kind of pointless? Since the mapping doesn't actually change anything, I think we could just return the original dict.

alpaca_trade_api/rest.py
```diff
@@ -884,6 +898,43 @@ def get_crypto_snapshot(self, symbol: str, exchange: str) -> SnapshotV2:
...
+    def get_news(self,
+                 symbol: Union[str, List[str]],
+                 start: Optional[str] = None,
+                 end: Optional[str] = None,
+                 limit: int = None,
+                 sort: Sort = Sort.Desc,
+                 include_content: bool = False,
+                 exclude_contentless: bool = False,
+                 ) -> NewsListV2:
+        news = list(self.get_news_iter(symbol,
```
gnvk, line 932: [minor] Please use key argument here too!

alpaca_trade_api/rest.py
```diff
@@ -884,6 +898,43 @@ def get_crypto_snapshot(self, symbol: str, exchange: str) -> SnapshotV2:
...
+                      start: Optional[str] = None,
+                      end: Optional[str] = None,
+                      limit: int = None,
+                      sort: Sort = Sort.Desc,
+                      include_content: bool = False,
+                      exclude_contentless: bool = False,
+                      raw=False) -> NewsIterator:
+        if isinstance(symbol, str):
+            symbol = [symbol]
+        news = self._data_get('news', symbol,
```
gnvk, line 912: I think it makes slightly more sense to have the `endpoint_base='news'` and `endpoint=''`. Also, please use key arguments!

alpaca_trade_api/rest.py
```diff
@@ -618,8 +630,10 @@ def _data_get(self,
             else:
                 path = f'/{endpoint_base}/{endpoint}'
                 data['symbols'] = ','.join(symbol_or_symbols)
+            if not endpoint_base or not symbol_or_symbols or not endpoint:
```
gnvk, line 633: This feels hacky. I suggest the following approach instead: 
- set `path` to `f'/{endpoint_base}`
- if `symbol_or_symbols` is str and not empty: add `f'/{symbol_or_symbols}'` to `path`
- if `endpoint` is not empty, add it as well

alpaca_trade_api/rest.py
```diff
@@ -618,8 +630,10 @@ def _data_get(self,
...
+                path = re.sub(r'(\/)\1+', r'\1', path)  # Remove extra slashes
             resp = self.data_get(path, data=data, api_version=api_version)
-            if isinstance(symbol_or_symbols, str):
+            if isinstance(symbol_or_symbols, str) or endpoint == "news":
```
gnvk, line 636: I don't think `endpoint == "news"` is the right abstraction here. I suggest the following approach instead:
- add something like an `Optional[bool]` `resp_grouped_by_symbol` argument to `_data_get` which if None can use the old behaviour (group if a list of symbols were provided, don't group if a single one)
- use that variable here
- for the news endpoint, set it to `False`

alpaca_trade_api/rest.py
```diff
@@ -884,6 +898,43 @@ def get_crypto_snapshot(self, symbol: str, exchange: str) -> SnapshotV2:
                              api_version='v1beta1')
         return self.response_wrapper(resp, SnapshotV2)
 
+    def get_news_iter(self,
+                      symbol: Union[str, List[str]],
```
gnvk, line 902: `symbol` should be Optional here and default to `None` to be consistent with the REST API.

alpaca_trade_api/rest.py
```diff
@@ -884,6 +898,43 @@ def get_crypto_snapshot(self, symbol: str, exchange: str) -> SnapshotV2:
...
+                              include_content=include_content,
+                              exclude_contentless=exclude_contentless)
+        for n in news:
+            if raw:
+                yield n
+            else:
+                yield self.response_wrapper(n, NewsV2)
+
+    def get_news(self,
+                 symbol: Union[str, List[str]],
```
gnvk, line 924: `symbol` should be Optional here and default to None to be consistent with the REST API.

alpaca_trade_api/rest.py
```diff
@@ -884,6 +898,43 @@ def get_crypto_snapshot(self, symbol: str, exchange: str) -> SnapshotV2:
                              api_version='v1beta1')
         return self.response_wrapper(resp, SnapshotV2)
 
+    def get_news_iter(self,
+                      symbol: Union[str, List[str]],
+                      start: Optional[str] = None,
+                      end: Optional[str] = None,
+                      limit: int = None,
```
gnvk, line 905: The limit handling in `_data_get` is incorrect unfortunately, because `DATA_V2_MAX_LIMIT` is not the actual max limit for the news API (it's 50 instead).

alpaca_trade_api/stream.py
```diff
@@ -771,12 +851,15 @@ async def stop_ws(self):
...
+        if self._news_ws:
+            await self._news_ws.stop_ws()
+
     def is_open(self):
         """
         Checks if either of the websockets is open
         :return:
         """
-        open_ws = self._trading_ws._ws or self._data_ws._ws or self._crypto_ws._ws  # noqa
+        open_ws = self._trading_ws._ws or self._data_ws._ws or self._crypto_ws._ws or self._news_ws # noqa
```
gnvk, line 862: [minor] This line is too long, please wrap it.

alpaca_trade_api/rest.py
ccnlui: Totally missed this. Thanks for catching

alpaca_trade_api/entity_v2.py
ccnlui: Removed

alpaca_trade_api/rest.py
ccnlui: Added `NEWS_MAX_LIMIT` and also exposing `total_limit` and `no_total_limit` parameters to `get_news()`

alpaca_trade_api/rest.py
```diff
@@ -600,27 +612,39 @@ def _data_get(self,
...
+            if isinstance(symbol_or_symbols, list) and symbol_or_symbols:
                 data['symbols'] = ','.join(symbol_or_symbols)
             resp = self.data_get(path, data=data, api_version=api_version)
-            if isinstance(symbol_or_symbols, str):
-                for item in resp.get(endpoint, []) or []:
+            if (
+                resp_grouped_by_symbol is False
+                or isinstance(symbol_or_symbols, str)
+            ):
+                # Try first key if empty endpoint
```
ccnlui, line 645: It feels like I'm still doing something hacky here...

`_data_get()` assumes we can use `endpoint` as a dict key to extract data from the response. My workaround is to use the first key available if `endpoint` is empty"

Let me know if you want to change this

ccnlui:
For `get_news()`, I'm using `NEWS_MAX_LIMIT = 50` as a default query limit to prevent fetching too many news articles. `no_total_limit` can be used to suppress this.

It also acts as a `page_limit` to minimize api calls for the client if `total_limit` is more than 50. This mimics behaviour of the other endpoints  like `get_trades()`

The difference is that the other endpoints don't have a default query limit... So `get_trades('AAPL')` would actually just fetch _everything_, 10000 data points at a time...

I wonder if I should add `total_limit` and `no_total_limit` to the other methods too, but it's a breaking change 🤔

alpaca_trade_api/rest.py
```diff
@@ -612,29 +612,38 @@ def _data_get(self,
                   symbol_or_symbols: Union[str, List[str]],
                   api_version: str = 'v2',
                   endpoint_base: str = 'stocks',
+                  resp_grouped_by_symbol: bool = None,
```
gnvk, line 615: [minor] the type hint should be `Optional[None]`

alpaca_trade_api/rest.py
```diff
@@ -612,29 +612,38 @@ def _data_get(self,
...
                   **kwargs):
         page_token = None
         total_items = 0
         limit = kwargs.get('limit')
+        page_limit = DATA_V2_MAX_LIMIT
+        if endpoint_base == "news":
+            page_limit = NEWS_MAX_LIMIT
```
gnvk, line 622: [minor] again, I would add this logic to the caller instead.

alpaca_trade_api/rest.py
```diff
@@ -899,40 +908,53 @@ def get_crypto_snapshot(self, symbol: str, exchange: str) -> SnapshotV2:
...
-                      symbol: Union[str, List[str]],
+                      symbol: Optional[Union[str, List[str]]] = None,
                       start: Optional[str] = None,
                       end: Optional[str] = None,
-                      limit: int = None,
                       sort: Sort = Sort.Desc,
                       include_content: bool = False,
                       exclude_contentless: bool = False,
+                      total_limit: int = None,
+                      no_total_limit: bool = False,
```
gnvk, line 918: A would just go with `limit: int = 10` instead for
- simplicity
- consistency with the other endpoints

alpaca_trade_api/rest.py
gnvk: Maybe a bit easier workaround would be to use `endpoint_base` if `endpoint` is empty. Both are kind of bad :D

alpaca_trade_api/rest.py
```diff
@@ -612,29 +612,38 @@ def _data_get(self,
...
                 data['symbols'] = ','.join(symbol_or_symbols)
-            if not endpoint_base or not symbol_or_symbols or not endpoint:
-                path = re.sub(r'(\/)\1+', r'\1', path)  # Remove extra slashes
             resp = self.data_get(path, data=data, api_version=api_version)
-            if isinstance(symbol_or_symbols, str) or endpoint == "news":
-                for item in resp.get(endpoint, []) or []:
+            if (
+                resp_grouped_by_symbol is False
+                or isinstance(symbol_or_symbols, str)
+            ):
```
gnvk, line 643: [minor] At the end of the day it's the same, but I feel it would be a bit more pythonic to check for None first outside the loop:
```python
if resp_grouped_by_symbol is None:
    resp_grouped_by_symbol = not isinstance(symbol_or_symbols, str)
```
and then simply use `resp_grouped_by_symbol` here as a condition.
```

File: alpaca_trade_api/rest.py (hunk 11 above)
Comment (ccnlui) [reply alpaca_trade_api/rest.py:615]: I assumed you meant `Optional[bool]`. Changed to that instead.

File: alpaca_trade_api/rest.py [hunk 15]
```diff
@@ -612,29 +612,38 @@ def _data_get(self,
...
-            if isinstance(symbol_or_symbols, str):
-                path = f'/{endpoint_base}/{symbol_or_symbols}/{endpoint}'
-            else:
-                path = f'/{endpoint_base}/{endpoint}'
+            path = f'/{endpoint_base}'
+            if isinstance(symbol_or_symbols, str) and symbol_or_symbols:
+                path += f'/{symbol_or_symbols}'
+            if endpoint:
+                path += f'/{endpoint}'
+            if isinstance(symbol_or_symbols, list) and symbol_or_symbols:
```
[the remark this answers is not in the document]
ccnlui: You're right. But I think this okay because the type hint is `symbol_or_symbols: Union[str, List[str]]`. So it's not breaking any promises. I'm leaving it if it's okay with you too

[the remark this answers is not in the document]
gnvk: Well, type hints are just hints in python. Technically, this is a breaking change.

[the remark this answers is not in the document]
ccnlui: Changed to previous condition

alpaca_trade_api/rest.py
ccnlui: added `page_limit` as a parameter to `_data_get()` and passing `NEWS_MAX_LIMIT` from the caller. 
Let me know if this is not what you had in mind

gnvk: Yes, this is exactly what I had in mind.

gnvk (approved):
LGTM. Thank you @ccnlui!

PR #558, opened 2022-01-19
Mark v1 marketdata endpoints deprecated

gnvk:
One additional note: the "overnight hold" example is still using `get_barset`, however updating it is not completely trivial.


camelpac (approved):
LGTM. regarding the overnight_hold example. when changes were made there was no support for the rest_async module so using `get_bars` was problematic. now the solution should be moving to the async rest module and stop using `get_barset`

Issue #412, opened 2021-04-05
Long short example is not working


andyliangdong:
HTTPRequest has exception 422 Client Error

haxdds:
Hi andyliangdong,

you may be receiving this error due to not having a premium data subscription.
The long-short example depends on the past 15 minutes of historical data, which is only available with a premium data subscription.

You'll know this is the case, if the error message contains a note about being unauthorized.

PR #560, opened 2022-01-28
Fixes HTTP 422 Issue on Long-Short Example

haxdds:
The start and end datetimes were changed to be in RFC-3339 Format.

### DIFF
--- a/examples/long-short.py
+++ b/examples/long-short.py
@@ -273,13 +273,23 @@ class LongShort:
   # Get the total price of the array of input stocks.
   def getTotalPrice(self, stocks, resp):
     totalPrice = 0
+
+    startTime = datetime.datetime.now(datetime.timezone.utc) - datetime.timedelta(minutes=1)
+    startTime = startTime.replace(second=0, microsecond=0).isoformat()
+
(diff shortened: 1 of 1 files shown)


ccnlui:
I don't see how this fixes any "HTTPRequest has exception 422 Client Error" errors in issue #412? How did you reproduce the issue?

In any case, if all you want is a timestamp in RFC-3339 Format, I would go with following:
```
In [10]: pd.Timestamp.utcnow()
Out[10]: Timestamp('2022-01-29 02:27:02.011808+0000', tz='UTC')

In [11]: str(pd.Timestamp.utcnow())
Out[11]: '2022-01-29 02:27:07.107777+00:00'
```

Much cleaner I would say, plus its pandas 🐼

haxdds:
@ccnlui 

When I ran the existing example. I received a 422 Error due to incorrect datetimes for start and end. You don't receive the error. 

And as for how we output the datetime. Although I agree pandas is more concise, I wrote it out in a way to make it more readable for people who may be looking at this as an example.

haxdds:
Removed timestamp changes.

PR contains changes for fixing incorrect access of bars list.

drew887 (approved):

PR #561, opened 2022-01-28
Python 3.10

ccnlui:
## Context.
Support python 3.10 by upgrading websockets library to 10.1 and replacing deprecated asyncio api methods.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,7 @@ jobs:
           name: run tests
           command: |
             pip install pandas==1.3.4 numpy==1.21.4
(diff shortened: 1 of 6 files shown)


ccnlui:
I added python 3.10 to circle CI. I also changed the install command from `python setup.py install` to `pip install .`
`python setup.py install` was having trouble picking the correct whl file for PyYAML, plus `pip install .` is the better/recommended way of installing packages anyway. So this would test the installation method documented in README.md more consistently.

haxdds (approved):
LGTM! Thank you!

PR #562, opened 2022-01-29
Updating overnight-hold to use the new data api (async get bars)

camelpac:
In this PR 
* we update the example code for overnight-hold to stop using the the old get_barset() endpoint and use the data ewndpoint with the asyncio wrapper. 
* Adding ProcessPoolExecutor for better performance during backtests
* optimized data access for backtesting
* Fix some logic in the live mode section
* Better data manipulation for the ratings calculation

### DIFF
--- a/examples/overnight_hold.py
+++ b/examples/overnight_hold.py
@@ -386,9 +386,9 @@ def run_live(api):
         cycle += 1
 
 
-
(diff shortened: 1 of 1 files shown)


gnvk (approved):
Thank you!

PR #577, opened 2022-02-22
empty handler check when unregister cancel errors + corrections

ccnlui:
## Context
Bug only exists for cancel errors and corrections because the `unregister_handler` method is called automatically when trades are unsubscribed.


gnvk (approved):
LGTM

ccnlui:
thanks!

PR #586, opened 2022-03-11
Send subscription requests in parts

noramehesz:


alpaca_trade_api/stream.py
```diff
@@ -63,6 +64,7 @@ def __init__(self,
         }
         self._name = 'data'
         self._should_run = True
+        self._maxSymbolsInMessage = 100  # should be change
```
ccnlui, line 67: I think we should stick with `_max_symbols_in_message` to be consistent

alpaca_trade_api/stream.py
```diff
@@ -191,17 +193,23 @@ def _subscribe(self, handler, symbols, handlers):
...
+                        msg = defaultdict(list)
+                        symbolsInMsg = 0
+                    msg[k].append(s)
+                    symbolsInMsg += 1
+        if msg:
+            await self._send_subscribe_msg(msg)
+
+    async def _send_subscribe_msg(self, msg):
+        msg['action'] = 'subscribe'
+        await self._ws.send(msgpack.packb(msg))
```
ccnlui, line 212: While it works, I think it's confusing to the users, because they will get back multiple subscription responses if the original subscription is broken down to multiple individual requests.

Can I suggest another solution, which sends the really big subscription message in smaller frames? Instead of sending multiple smaller subscription messages.
```python3
msg['action'] = 'subscribe'
bs = msgpack.packb(msg)
frames = ( bs[i:i+self._max_frame_size] for i in range(0, len(bs), self._max_frame_size) )
await self._ws.send(frames)
```

I got this from the [documentation](https://websockets.readthedocs.io/en/stable/reference/client.html#websockets.client.WebSocketClientProtocol.send): 
>[send()](https://websockets.readthedocs.io/en/stable/reference/client.html#websockets.client.WebSocketClientProtocol.send) also accepts an iterable or an asynchronous iterable of strings, bytestrings, or bytes-like objects to enable [fragmentation](https://www.rfc-editor.org/rfc/rfc6455.html#section-5.4). Each item is treated as a message fragment and sent in its own frame. All items must be of the same type, or else [send()](https://websockets.readthedocs.io/en/stable/reference/client.html#websockets.client.WebSocketClientProtocol.send) will raise a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) and the connection will be closed.

alpaca_trade_api/stream.py
```diff
@@ -64,7 +64,7 @@ def __init__(self,
         }
         self._name = 'data'
         self._should_run = True
-        self._maxSymbolsInMessage = 100  # should be change
+        self._max_frame_size = 128  # should be change
```
ccnlui, line 67: 128 bytes is probably too little. I would put `32768`. But maybe ask @gnvk too 🤔

ccnlui (approved):
LGTM. I think it can be merged after `_max_frame_size` is adjusted

PR #590, opened 2022-03-18
Remove marketdata v1 functions

gnvk:


alpaca_trade_api/stream.py
```diff
@@ -1,8 +1,3 @@
-"""
-Stream V2
```
ccnlui, line 2: Should we keep `Stream V2` here? I'm fine with remove the rest of the comments, but I wonder if it's better to state this is V2 explicitly.

ccnlui:
I think this paragraph is no longer relevant, we don't have this flag in `Stream`

https://github.com/alpacahq/alpaca-trade-api-python/blob/c0384646329535faeeb3fb82dbf2200d4cd6f488/README.md?plain=1#L382-L385

ccnlui:
This I'm not sure about. Is this still relevant?
https://github.com/alpacahq/alpaca-trade-api-python/blob/c0384646329535faeeb3fb82dbf2200d4cd6f488/README.md?plain=1#L409-L412

ccnlui:
Replace `StreamConn` with `Stream`
https://github.com/alpacahq/alpaca-trade-api-python/blob/c0384646329535faeeb3fb82dbf2200d4cd6f488/README.md?plain=1#L416

ccnlui (approved):
Just a few comments regarding documentation. LGTM otherwise 👍

alpaca_trade_api/stream.py
gnvk: Actually it isn't. Crypto is in v1beta1, so is the news stream.

PR #591, opened 2022-03-18
Increase ping timeout

gnvk:


alpaca_trade_api/stream.py
```diff
@@ -67,7 +67,9 @@ def __init__(self,
     async def _connect(self):
         self._ws = await websockets.connect(
             self._endpoint,
-            extra_headers={'Content-Type': 'application/msgpack'})
+            extra_headers={'Content-Type': 'application/msgpack'},
+            ping_timeout=180,
+        )
```
drew887, line 72: This should probably also set `ping_interval` to 10? Looks like the default is 20 seconds

drew887 (changes requested):

alpaca_trade_api/stream.py
gnvk: That doesn't really matter, it doesn't have to be symmetrical.

PR #599, opened 2022-03-31
Revert "stream crypto orderbook"

ccnlui:
Reverts alpacahq/alpaca-trade-api-python#597 because it is not on production yet.


ccnlui:
Crypto orderbook is now in production. I removed this "revert" commit to preserve a cleaner history. Sorry for the confusion.

Issue #607, opened 2022-04-21
[Question]: list_orders() Quantity


sshcli:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

This is not a bug, it is a question.
I can't find any documentation or example about this.

I know how to show a list of orders:
```
api = alpaca.REST(api_key, api_secret, base_url, api_version='v2')  
ORDERS = api.list_orders()
```

But what I don't know, is how to get the quantity of orders I have open right now.
I don't have the order id, I just have the symbol

**For example:** 
I want to know the exact qty of open orders I have to sell TSLA to avoid submitting another order.

Thanks in advance for your response.

### Expected Behavior

Get orders qty 

### SDK Version I encountered this issue in

Alpaca-trade-api- == 2.0.0

### Steps To Reproduce

```markdown
Try to get the quantity of open orders for a given symbol (Ex: TSLA)
```

### Filled out the Steps to Reproduce section?

- [X] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

drew887:
Hey @sshcli sorry about the issues with the bug template there, we're going to add a template for generalized questions. 

If you want to get the list of open orders for a specific set of symbols `list_orders` takes a `symbols` parameter that lets you filter to a list of specific symbols. 

For example:
```python
api = alpaca.REST(api_key, api_secret, base_url, api_version='v2')  
ORDERS = api.list_orders(symbols=['TSLA'])
```
Is that what you were looking for ?

sshcli:
> Hey @sshcli sorry about the issues with the bug template there, we're going to add a template for generalized questions.

Great, no problem.

> Is that what you were looking for ?

No. What I want to get is the quantity of all open orders. the result must be a integer number. For example: 3

drew887:
Ah sorry, did you mean the sum of the quantity field of all open orders? Not just the number of orders?

If that's the case then that's not really something we need to provide direct capabilities too into the SDK imo, as it would be common/easy enough to do it with general python functionality. IE reducing over the list of returned orders or using a for loop would be the first things that come to mind.

For example does something like :
```python
from functools import reduce

api = alpaca.REST(api_key, api_secret, base_url, api_version='v2')  
orders = api.list_orders(symbols=['TSLA'])

total_qty = reduce(
      lambda result, order: result + int(order.qty),
      orders,
[...]

PR #608, opened 2022-04-21
Add issue template for general SDK questions

drew887:
See #607 for an example case of why we want a generalized question template


sshcli:
> See #607 for an example case of why we want a generalized question template

@drew887 I agree with you, this is really necessary. I vote for a merge

haxdds (approved):

PR #615, opened 2022-05-03
Updated Stream Instructions for `sip` parameter in README

hinchcliffz:
I updated the Stream instructions to be more clear.
- They had an unneeded URL in the `README.md` updated it to reflect the proper way to call it.
- They had SIP parameter in all capitals in the instructions as well. Fixed this to reflect the proper way to call it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ async def quote_callback(q):
 # Initiate Class Instance
 stream = Stream(<ALPACA_API_KEY>,
                 <ALPACA_SECRET_KEY>,
-                base_url=URL('https://paper-api.alpaca.markets'),
+                base_url='https://paper-api.alpaca.markets',
                 data_feed='iex')  # <- replace to SIP if you have PRO subscription
 
 # subscribing to event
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -249,8 +249,8 @@ async def quote_callback(q):
 # Initiate Class Instance
 stream = Stream(<ALPACA_API_KEY>,
                 <ALPACA_SECRET_KEY>,
-                base_url=URL('https://paper-api.alpaca.markets'),
-                data_feed='iex')  # <- replace to SIP if you have PRO subscription
+                base_url='https://paper-api.alpaca.markets',
+                data_feed='iex')  # <- replace to 'sip' if you have PRO subscription
```
drew887, line 253: Can you please add this comment to the second example below as well please?

drew887 (changes requested):
> They had an unneeded URL in the `README.md` updated it to reflect the proper way to call it.

I wouldn't say that you changed it to the "proper" way to call it. We do annotate the `base_url` parameter as a URL type so having it be passed as the URL like it already was would technically be the "proper" way to call it; and it also ensures that what you're passing in is a valid url. 

I think I'd rather we keep these examples to have the URL parts there just to help ease any issues that might come up with users accidentally making typos or the like. 

If you'd like us to be explicit about the fact that technically you could pass in a raw str value here, then we should update the type annotations on the `Stream` and `TradingStream` classes to show it can be a `Union[URL, str]` in this PR :thinking:

README.md
hinchcliffz: Sure @drew887 just added it.

hinchcliffz:
That makes a lot more sense @drew887. I added the URL back in and added the additional comment in the `README.md`.

drew887 (approved):

PR #620, opened 2022-05-13
apache.org is now using https protocol

sshcli:


drew887:
It looks like [the official apache version](https://www.apache.org/licenses/LICENSE-2.0) still uses http

PR #628, opened 2022-06-15
Fix missing URL import for the Stream class example in README

hiohiohio:
Add URL import to solve below error for the Stream class example in README
```
NameError: name 'URL' is not defined
```


haxdds (approved):
Thank you!

PR #632, opened 2022-06-23
Add optional asof for historical marketdata endpoints

gnvk:


noramehesz (approved):
Looks good to me! 👍

benlevitas:
It would be nice if asof also supported different adjusted prices. Meaning that if we were to look back a year at adjusted prices with no asof we might get a price that has been adjusted 4 times, while looking back a year with an asof that was 9 months ago we will get a price that has been adjusted once.

sshcli:
> It would be nice if asof also supported different adjusted prices. 

@benlevitas I think you should create a feature request in the new official repo:
https://github.com/alpacahq/alpaca-py

PR #672, opened 2022-12-19
Use a less restrictive requirement specification for `aiohttp`.

### LINKED ISSUE #671: [Bug]: Fails to install with Python 3.11 due to `aiohttp==3.8.1` dependency
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using Python 3.11, `pip install alpaca-trade-api~=2.0` fails with the following error for the `aiohttp` dependency:
```
aiohttp/_websocket.c:198:12: fatal error: 'longintrepr.h' file not found
  #include "longintrepr.h"
           ^~~~~~~~~~~~~~~
1 error generated.
error: command '/usr/bin/clang' failed with exit code 1
----------------------------------------
ERROR: Failed building wheel for aiohttp
```

### Expected Behavior

`pip install alpaca-trade-api~=2.0` with Python 3.11 should succeed.

### SDK Version I encountered this issue in

2.3.0

### Steps To Reproduce

```markdown
1. Use Python 3.11.
2. Run `pip install alpaca-trade-api~=2.0`.
```

### Filled out the Steps to Reproduce section?

- [X] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

### Anything else?

This [issue](https://github.com/aio-libs/aiohttp/issues/6600) has already been fixed starting with `aiohttp` version 3.8.3. However, the `alpaca-trade-api` package [pins](https://github.com/alpacahq/alpaca-trade-api-python/blob/bf02b2b870de5c9961b867908698faed0c738abb/requirements/requirements.txt#L8) its `aiohttp` dependency to version 3.8.1. We should use a less restrictive requirement specification so that future `aiohttp` fixes are brought in automatically.

fumoboy007:
Fixes #671.

We should use a less restrictive requirement specification so that future `aiohttp` fixes are brought in automatically. `aiohttp~=3.8` will enable automatic upgrading to both new patch versions and new minor versions while disabling automatic upgrading to new major versions, which may have breaking API changes.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -5,6 +5,6 @@ urllib3>1.24,<2
 websocket-client>=0.56.0,<2
 websockets>=9.0,<11
 msgpack==1.0.3
-aiohttp==3.8.1
+aiohttp~=3.8
 PyYAML==6.0
 deprecation==2.1.0


fumoboy007:
Closing in favor of #670.

PR #691, opened 2023-03-27
Fix streaming news article event dispatching

fumoboy007:
News article events are different from other types of events (e.g. stock trades) because a single news article could be applicable to multiple symbols. For that reason, the streamed event does not have the usual `S` field that contains a single symbol; instead, the event has a `symbols` field that contains a list of symbols.

#555 mistakenly assumes that the event has an `S` field. This commit changes the dispatching logic to instead use the `symbols` field.

(There is a similar [commit](https://github.com/alpacahq/alpaca-trade-api-csharp/commit/be67b62d0225551c828c8a1ed5985a7e760c8acd) for the C# library.)

PR #693, opened 2023-03-28
Only call the news handler once for a single message

gnvk:
Caused by #691


fumoboy007:
Isn’t this not the desired behavior?

Consider the case where a caller registers one handler for one symbol and a different handler for another symbol. If there is a news article that is related to both symbols, only one of the handlers would be called due to this change.

fumoboy007:
I think we should still try to call all of the handlers but avoid calling the same handler twice.

gnvk:
@fumoboy007 👍 https://github.com/alpacahq/alpaca-trade-api-python/pull/694

PR #694, opened 2023-03-31
Call each different news handler once

gnvk:


alpaca_trade_api/stream.py
```diff
@@ -573,15 +573,14 @@ async def _dispatch(self, msg):
...
                 symbols.append('*')
 
+            handlers = set()
             for symbol in symbols:
-                handler = self._handlers['news'].get(symbol)
-                if handler is None:
-                    handler = self._handlers['news'].get('*')
-
+                handler = self._handlers['news'].get(
+                    symbol, self._handlers['news'].get('*'))
```
fumoboy007, line 579: Minor efficiency issue: This way of writing the statement will always evaluate `self._handlers['news'].get('*')` even if there is a non-wildcard handler, which is why I split the statement into two via `if handler is None`.

PR #745, opened 2024-01-11
Support Python 3.12

hiohiohio:
- support python 3.12
- alpaca-trade-api-python is deprecated  and recommend to use [alpaca-py](https://github.com/alpacahq/alpaca-py)


gnvk (approved):
LGTM

